PR #15, opened 2011-12-06
tools/get_status: only look at ur-agent if it exists, get kvm and kvm-cmd statuses

richlowe:
Mostly, only get the git status of ur-agent if one actually exists (which it won't outside of, I presume, Joyent), thus gagging a warning from 'gmake live'.

While here, also pull the status of illumos-kvm and illumos-kvm-cmd, since they're there and it seemed worth doing.


sjorge:
@richlowe Is this still relevant? If not can this be closed, if so it looks like it needs a rebase.

jlevon:
This was long since made irrelevant by OS-2060

PR #58, opened 2012-02-10
Support for VNC config, media control and vga config

essele:
1. Adds support for:
   vmadm media <uuid> eject <device> [force]
   vmadm media <uuid> change <device> <target>      (target is relative to zone root)
2. Adds configurable qemu vga settings (defaults to cirrus as before)
3. Adds config options for vnc:
   vnc: on, off, or secure    -- defaults to on (as before), off disables it, secure requires a password.
   vnc_port -- allows you to specify a port to listen on that's fixed, default is 0, which is random, as before.
   vnc_password -- the password used if 'secure' mode is selected.

I've updated both the vmadm and vmadmd manual pages, and I've tried to keep the coding style the same,
although it may need the odd tweak.

I've probably done some silly things around error checking, but it gets quite nested by the time you've gone
from vmadm, via VM.js, to vmadmd and then talking to the qmp port.

On the vnc side, I had to make the vnc connection re-establish since it gets dropped if you get the password
wrong … so I call spawnVNC from the connection handler within spawnVNC, I _think_ this should be ok given
it should only happen on connection drops, but it might be worth a bit of investigation.

At the moment I haven't included vnc_password in the vnc info output since it's available with "get".


sjorge:
I don't think the adding of cdroms/swapping iso ever got added? using the nc -U to the qemu-monitor socket still the prefered way?

PR #114, opened 2012-08-25
imgadm show does not work with manually installed images

MerlinDMC:
`imgadm show` is just checking the cache available from the repository server. If you try to get info on an image installed via `imgadm install` it just exists with

```
Error: not found
```

This patch searches for already installed images if the uuid is not in the cache.


xen0l:
This PR is also very helpful if somebody has to work with local images on a daily basis. @MerlinDMC do you think you could resolve the conflicts?

MerlinDMC:
@xen0l I can try but since imgadm did divert a good bit from the old codebase it will be a little more work. Will put it on my list but if someone else wants to do it feel free to start.

PR #130, opened 2012-10-15
imgadm-repair destroys the imgadm database

### LINKED ISSUE #129: imgadm-repair seems to be broken
on platform version joyent_20121004T212912Z

```
[root@00-0c-29-18-ec-10 ~]# /usr/img/sbin/imgadm-repair
Checking for manifest with UUID: $i
$i - manifest was found. Copying to $DBDIR
json: error: input is not JSON: Unexpected '$' at line 1, column 1:
        ${manifest}
        ^
$i - urn: $urn
[root@00-0c-29-18-ec-10 ~]# imgadm info 01b2c898-945f-11e1-a523-af1afbe22822

undefined:1
${manifest}
^
SyntaxError: Unexpected token $
    at Object.parse (native)
    at db.load (/usr/img/node_modules/db.js:182:37)
    at fs.readFile (fs.js:176:14)
    at Object.oncomplete (fs.js:297:15)
```

funny about that is that even with the error the cachefile gets updated:
```
[root@00-0c-29-18-ec-10 ~]# cat /var/db/imgadm/01b2c898-945f-11e1-a523-af1afbe22822.json
${manifest}
```

now `imgadm list` ect. is not working anymore

MerlinDMC:
the variables in single-quotes don't get replaced so imgadm-repair writes `${manifest}` into all json files which breaks everything else that should work with the file.

imgadm-repair tried to update locally imported manifest which are not found in the cache file - that would save an empty manifest file.

this pull request fixes #129

### DIFF
--- a/src/img/sbin/imgadm-repair
+++ b/src/img/sbin/imgadm-repair
@@ -51,7 +51,7 @@ for i in ${zvols[@]} ; do
     echo "Checking for manifest with UUID: $i"
     manifest=$(imgadm show $i)
 
-    if [[ -n $manifest ]] ; then
+    if [[ "$?" == "0" && -n $manifest ]] ; then
         echo "$i - manifest was found. Copying to $DBDIR"
         echo "${manifest}" > $DBDIR/$i.json
         urn=$(cat $DBDIR/$i.json | json urn)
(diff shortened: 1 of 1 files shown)


bahamas10:
While this was clearly a bug, it looks like this script was removed in https://github.com/joyent/smartos-live/commit/f729853997f7607baa7ed4431985ebad3263b138

@MerlinDMC is this still relevant?

MerlinDMC:
@bahamas10 if the script is gone this somewhat counts as "fixed" for me.

Issue #120, opened 2012-09-25
imgadm does not work behind proxies


khushil:
When using the downloadable image on local machines, at times we are forced behind corporate firewalls and have to use the export vars http_proxy / https_proxy to access the outside world.

When setting http_proxy and https_proxy imgadm seems to ignore these settings and try a direct connect which will error in net.js with ETIMEDOUT.

imgadm should support http_proxy and https_proxy.

trentm:
This is fixed now in #master. I'm sorry that it took this long.

bapt:
Usually applications accepts both HTTP_PROXY and http_proxy. imgadm only supports lower case. Would be nice to also accept upper case version

PR #142, opened 2012-11-12
make imgadm to use http(s)_proxy environment variables

MerlinDMC:
This request changes imgadm to use curl for downloading manifests from dataset servers.

Curl respects http(s)_proxy environment variables during download so imgadm can also be used in behind proxies.

This fixes issue #120

### DIFF
--- a/src/img/node_modules/imgadm.js
+++ b/src/img/node_modules/imgadm.js
@@ -58,8 +58,6 @@ var common = require('./common');
 var db = require('./db');
 var dns = require('dns');
 var fs = require('fs');
-var http = require('http');
-var https = require('https');
 var log = common.log;
 var path = require('path');
 var spawn = require('child_process').spawn;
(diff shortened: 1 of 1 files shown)


trentm:
This should be fixed in smartos-live#master now. I'm sorry this sat idle for so long.

AlainODea:
Thanks @trentm !  This is great news.

src/img/node_modules/imgadm.js
```diff
@@ -314,7 +253,8 @@ var cacheUpdate = function (callback) {
 
         // SmartOS DNS is disabled by default
         dns.resolve4(options.hostname, function (err, addresses) {
-            var client;
+            var args = ['--insecure'];
```
bahamat: If nothing else, the existing code will validate certificates by default and I think it's entirely inappropriate to disregard certificate errors.

PR #198, opened 2013-04-04
Root check in smf_import

szaydel:
Looks like there was an old TODO to add a root check to script, so I thought I could quickly add it.

### DIFF
--- a/src/sysinfo
+++ b/src/sysinfo
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2019 Joyent, Inc.
+# Copyright 2020 Joyent, Inc.
 #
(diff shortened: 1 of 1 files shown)


bahamas10:
A bunch of commits look like they have been pushed to this branch, rendering the diff unreadable.

nshalman:
@bahamas10 I think the intent of the PR was just that first commit ( https://github.com/szaydel/smartos-live/commit/cd6a396e88239b53bbe1d94f966c243ec6234588 )

szaydel:
This is very old, which may mean it is no longer relevant. It is a tiny snippet, so if still relevant, I am sure we can fix it so it is applied after all the commits that have happened since this original one.

PR #242, opened 2013-07-27
OS-1 update to json 6.0.0 for jsontool inplace editing of json files

jacques:


bahamas10:
The goal now is to upgrade `json` to the latest version available, which at the time of this writing is `9.0.4`.  There is a breaking change that is preventing us from doing this currently noted here https://github.com/trentm/json/blob/master/CHANGES.md#json-900 under "backwards incompatibility".

The idea is to get all of the smartos bits updated to use the proper `this.` notation, or to bring back the old behavior (see https://github.com/trentm/json/pull/93/files).  Regardless, the end goal will be to upgrade JSON to the latest in the platform.

PR #313, opened 2014-03-10
fixed separateCommas() on empty string

dn0:
Running String.split() on empty string produces an array with one -- empty string -- element (`[ "" ]`). This leads to the following problem:

Setting the `nics.*.allowed_ips` to empty array with `vmadm update` will successfully set the zone property to empty string, but the output of `vmadm get` is not reusable and for example a consecutive `vmadm send/receive` will fail with _invalid allowed_ips_.

### DIFF
--- a/src/vm/node_modules/utils.js
+++ b/src/vm/node_modules/utils.js
@@ -244,6 +244,10 @@ function rtrim(str, chars)
 
 function separateCommas(str)
 {
+    if (str.length === 0) {
+        return [];
+    }
+
     return str.split(',');
 }
 


bahamas10:
The function below it probably also needs the same treatment, `separateCommasAndNumberify`.  In fact, it probably makes sense for `separateCommasAndNumberify` to use `separateCommas` and then numberify the results.

This change makes sense - It'll just need to be tested to ensure nothing breaks as a result of it.

xen0l:
So, what has to be done for this PR to be accepted?

bahamas10:
@dn0 @xen0l The function below, `separateCommasAndNumberify` should also get this treatment... I would recommend updating it to look like:

``` js
function separateCommasAndNumberify(str)
{
    return separateCommas(str).map(function (val) {
        return numberify(val);
    });
}
```

bahamas10:
also, i'm unable to reproduce the current issue (re: `nics.*.allowed_ips`), can you post example command output so I can test it as well? thanks.

dn0:
@bahamas10 The idea is to have a proper way to clear the `nics.*.allowed_ips` property (and other array properties, such as `nics.*.blocked_outgoing_ports`). After setting the attribute to empty array the subsequent `vmadm get` will return a json with an empty string in the `nics.*.allowed_ips` array:

```json
# vmadm get 1043817b-e2fd-46bd-8b81-fb7a3a015bea | json nics.0
{
  "interface": "net0",
  "mac": "c2:a1:ea:89:32:9f",
  "allowed_ips": [
    "172.18.0.241",
    "172.18.0.242"
  ]
   ....
}
```

```json
# cat test.json 
{
        "uuid": "1043817b-e2fd-46bd-8b81-fb7a3a015bea",
        "update_nics": [
                {
                        "mac": "c2:a1:ea:89:32:9f",
                        "allowed_ips": []
                }
        ]
}
```

```json
# cat test.json | vmadm update 1043817b-e2fd-46bd-8b81-fb7a3a015bea
Successfully updated VM 1043817b-e2fd-46bd-8b81-fb7a3a015bea

# vmadm get 1043817b-e2fd-46bd-8b81-fb7a3a015bea | json nics.0
{
  "interface": "net0",
  "mac": "c2:a1:ea:89:32:9f",
  "allowed_ips": [
    ""
  ],
   ...
}

# vmadm send 1043817b-e2fd-46bd-8b81-fb7a3a015bea  | ssh root@new-world vmadm receive
invalid allowed_ips: 

Tested on 20150226T162528Z
```

Anyway, this PR is really old and as you've mentioned it should be fixed in a more global way. Since I don't really remember the vmadm code and you now a better way to fix it, feel free to close this issue.

PR #327, opened 2014-06-11
build: git --depth=1 sample.configure.smartos

burlito:
just added --depth=1 to every git clone in sample.configure.smartos

I just don't think there is any reason to clone full repository when we need just last snapshot.


bahamas10:
I'm -1 on this.. I prefer to always have the full history when doing development.  I could see potentially making this a non-default option to use a shallow clone however.

Issue #118, opened 2012-09-14
adding some services to joyent-minimal (not enabled or imported by default)


nshalman:
We have a use case that seems like a great fit for joyent-minimal if the following services could be enabled by us after the zone boots:

svc:/network/routing-setup:default
svc:/network/ipv4-forwarding:default
svc:/network/routing/route:default
svc:/network/ipfilter:default

rmustacc:
Resolved in 7d0701f.

PR #339, opened 2014-08-13
make joyent-minimal brand more flexible

nshalman:
fixes joyent/smartos-live#118

joyent-minimal and joyent brands now share the
/zones/manifests/joyent lofs mount

This is safe because the minimal brand overwrites the SMF repository of
whatever image is being used with its own minimal seed repository which
disables early-manifest-import and manifest-import

This change allows a joyent-minimal zone to import any other service
available to the joyent brand without maintaining a complicated brand
manifests file

### DIFF
--- a/overlay/generic/lib/svc/method/svc-zones
+++ b/overlay/generic/lib/svc/method/svc-zones
@@ -137,6 +137,8 @@ setup_manifests()
 	do
 		brand=`basename $i`
 		[[ ! -f /usr/lib/brand/$brand/manifests ]] && continue
+		# joyent-minimal uses /zones/manifests/joyent too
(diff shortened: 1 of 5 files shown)


nshalman:
I followed https://blog.shalman.org/running-sdc-coal-on-smartos/ but before booting I replaced the platform with my own platform including both this change and the Delphix vioif change (see https://github.com/joyent/sdc-manta/pull/2 for a necessary workaround to get CoaL to cooperate with vioif).

I was able to get SDC and Manta installed.
I was able to put and get an object and run a job over it.
I also installed madtom and it showed everything was green.

Does that count as enough testing for this change?

rmustacc:
Resolved in 7d0701f.

nshalman:
Thanks @rmustacc!

PR #381, opened 2014-12-20
fix vmtype to make more sense, add ostype

sjorge:
some cleanups to the vmadm man page

### DIFF
--- a/src/vm/man/vmadm.1m.md
+++ b/src/vm/man/vmadm.1m.md
@@ -501,7 +501,11 @@ tab-complete UUIDs rather than having to type them out for every command.
 
         type -- type of the properties value.
 
-        vmtype -- types of VM (KVM, LX, OS) for which this property applies.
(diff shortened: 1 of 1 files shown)


joshwilsdon:
Why do you think that this change is necessary? I'm not sure that I agree that this change makes more sense.

From my perspective the "type" field here is intended to expose which type of virtualization / emulation we're using. All of the current "OS" type brands use native OS virtualization where all syscalls are untranslated. The current "KVM" type uses the kvm hardware virtualization and is obviously emulating everything at the hardware level. The "LX" type was set here because the virtualization is not directly native since we're emulating *Linux* syscalls for applications in this zone.

sjorge:
The intent is to make it more clear that all stuff that applies to OS type also applies to LX but not the other way around.

Current types are KVM, OS, and LX but OS and LX are basically the same, except that lx takes a few extra's like kernel version.

PR #395, opened 2015-02-06
OS-3810: tar doesn't properly wait for its children

nshalman:
Workaround for build systems with a broken illumos tar.


trentm:
holding on this for now. might instead put a guard for min platform in that build script :)

trentm:
merged before the builds for this bi-weekly build. Thanks!  commit 338cee7708d19f072cb001e4686ce753505a8bdd after rebasing

PR #403, opened 2015-02-22
Update ca-bundle.crt to the latest version

xen0l:
The curlCA version shipped with SmartOS is from 2010, which is almost 5 years old. We should update it to the latest possible version. I added the version from https://raw.githubusercontent.com/bagder/ca-bundle/master/ca-bundle.crt (last available according to http://curl.haxx.se/docs/caextract.html).


lloydde:
Became #405 "Stop delivering curlCA".

rmustacc:
To clarify, this file wasn't actually being used by curl or anything else in the platform, hence why this was closed.

PR #404, opened 2015-02-27
Fix /etc/profile in generic overlay after 3842

xen0l:
After illumos removed machid commands, I can see "bash: /bin/i386: No such file or directory" in freshly built smartos. This is caused by missing changes to /etc/profile in smartos-live. This PR fixes that.


rmustacc:
Stock illumos /etc/profile is also broken. I'm going to be taking this up immediately with them and will figure out the best path forward, it may be that 3824 is backed out.

rmustacc:
Resolved in eba928b.

PR #405, opened 2015-02-27
Stop delivering curlCA

xen0l:


rmustacc:
As this isn't actually used by anything, it just gives us a false sense of purpose. Removing it makes sense.

rmustacc:
Resolved in ee3fb13.

Issue #407, opened 2015-03-10
imgadm create fails to create image from KVM instance without image_uuid


xen0l:
I tried to create image from custom installed CentOS 6 machine via imgadm create. However, imgadm failed with the following error: https://gist.github.com/xen0l/b89b04d79eb44a608e05

vmadm get output: https://gist.github.com/xen0l/b82f732e768a46a24c98

I tried it on KVM instances with cloned disk and it worked. imgadm create shouldn't expect image_uuid to be present on disks or is there any reason?

trentm:
Fixed in https://github.com/joyent/smartos-live/pull/412  Thanks!

PR #412, opened 2015-03-11
imgadm create should work for non-cloned KVM VMs

xen0l:
Fix for #407

With this changeset, imgadm create works on non-clone KVM VMs and allows us to create image from freshly installed virtual machine. This was discussed with @trentm.


trentm:
merged in the commits above.

Issue #435, opened 2015-03-31
sdc uploads not found on SmartOS


RSully:
Email from cron on SmartOS:

> find: cannot open /var/log/sdc/upload/: No such file or directory

ryancnelson:
this is in 
https://github.com/joyent/smartos-live/blob/f311853cd8b896183302122acf18bb39ca6cdd33/overlay/generic/etc/cron.d/crontabs/root

... that crontab entry assumes the existence of directory /var/log/sdc/upload
we should probably put in a test 

similarly, but already-fixed, in 
https://github.com/joyent/sdc-platform/blob/8ba19f62f7c3e74f96303d0570ca403d7622f5ff/overlay/etc/cron.d/crontabs/root 
... there's a test, then cron if the test passes, like:
" [ -x /opt/smartdc/bin/sdc-phonehome ] && /opt/smartdc/bin/sdc-phonehome "

RSully:
Crontab file also exists in smartos-live besides sdc-platform.

jclulow:
Fixed via #438

PR #438, opened 2015-04-05
Add conditional to SDC log upload

RSully:
This is SDC specific and causes problems on vanilla SmartOS.

This is also implemented in SDC-specific crontab (see https://github.com/joyent/sdc-platform/blob/8ba19f62f7c3e74f96303d0570ca403d7622f5ff/overlay/etc/cron.d/crontabs/root).

It may be more appropriate to remove this line completely from smartos-live. 
Since that is a potentially breaking change I did not go that route for this pull-request.

Relevant Github issue: #435


jclulow:
Merged in be5fc42716a719f5cb60a9559cea470beea0b7eb -- Thanks!

PR #439, opened 2015-04-07
Want improved installer

nshalman:
This is a cleaned up version of work originally done by @rmustacc.
@sjorge and I have both done a fair amount of testing of this code in my hourly builds.

### DIFF
--- a/overlay/generic/smartdc/lib/smartos_prompt_config.sh
+++ b/overlay/generic/smartdc/lib/smartos_prompt_config.sh
@@ -11,16 +11,33 @@
 
 # XXX - TODO
 # - if $ntp_hosts == "local", configure ntp for no external time source
-# - try to figure out why ^C doesn't intr when running under SMF
+
+exec 4>>/var/log/prompt-config.log
+echo "=== Starting prompt-config on $(tty) at $(date) ===" >&4
+export PS4='[\D{%FT%TZ}] $(tty): ${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+export BASH_XTRACEFD=4
+set -o xtrace
+
+if [[ -z $(/bin/bootparams | grep "^smartos=true") ]]; then
+	echo "This script should only be run on SmartOS"
+	exit 1
+fi
 
 PATH=/usr/sbin:/usr/bin
 export PATH
+
+. /lib/svc/share/smf_include.sh
+
 . /lib/sdc/config.sh
 load_sdc_sysinfo
 load_sdc_config
 
+# ERRORS (for getanswer's errno)
+ENOTFOUND=1
+EBADJSON=2
+EUNKNOWN=3
+
 # Defaults
-datacenter_headnode_id=0
 mail_to="root@localhost"
 ntp_hosts="0.smartos.pool.ntp.org"
 dns_resolver1="8.8.8.8"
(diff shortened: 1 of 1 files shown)


rmustacc:
Probably the most useful thing here is to go through the test matrix and make sure that we've exercised all of the different paths, especially around things like dealing with existing zpools, custom creation, shelling out, causing failure, etc.

nshalman:
I did a bunch of tests with and without existing pools, testing of the custom pool creation and hitting ctrl-c, etc. which was how I found the issues fixed in my old repo in commit https://github.com/nshalman/sdc-platform/commit/7ab21aeca82cd9f98c51c87e8a48a2372e71eeeb

If you have an existing list of tests you want me to run I'm very happy to do that.
If you don't have an existing lists of tests please let me know.

rmustacc:
I don't have an explicit one of the top of my head. I can try to put it together, but putting together a full list of what explicitly has been done so far will help.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -30,73 +47,137 @@ dns_resolver2="8.8.4.4"
...
+	echo "Bringing up a shell.  When you are done in the shell hit ^D to"
+	echo "return to the system configuration tool."
+	echo
+
+	/usr/bin/bash
+
+	echo
+	echo "Resuming the system configuration tool."
+	echo
+	printf "$prmpt_str"
```
bahamas10: `printf` will interpret special strings like `%s`, `%c`, etc. inside `$prmpt_str`.  you should use

``` bash
printf '%s' "$prmpt_str"
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -30,73 +47,137 @@ dns_resolver2="8.8.4.4"
...
-	max_d=$(expr $fmax - 1)
-	max_host="$max_a.$max_b.$max_c.$max_d"
-
-	net_a=$(($net_a & $1))
-	net_b=$(($net_b & $2))
-	net_c=$(($net_c & $3))
-	net_d=$(($net_d & $4))
-
-	host_addr=$(($addr_d & ~$4))
+	num_a=$(($1 << 24))
```
bahamas10: in some functions `local` is used, in this one it is not.  all of these variables will be made global.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -30,73 +47,137 @@ dns_resolver2="8.8.4.4"
...
-# 7 in the 10.99.99.7 example above).  Also, max_host stores the max. host
-# number (e.g. 10.99.99.254 in the example above).
-#
-ip_netmask_to_network()
+ip_to_num()
 {
 	IP=$1
-	NETMASK=$2
 
 	OLDIFS=$IFS
```
bahamas10: you can skip the modification of `$IFS` globally (even though it is still used up by using some bash techniques)

``` bash
ip_to_num() {
  local ip=$1
  local a b c d
  IFS=. read a b c d <<< "$ip"
  ((a <<= 24))
  ((b <<= 16))
  ((c <<= 8))
  echo "$((a + b + c + d))"
}

num=$(ip_to_num 1.2.3.4)
```

notice that `$IFS` is only modified for the `read` builtin, and the function prints the output, so it can be captured by the caller (like a small program)

konobi: of course IFS would need to be reset back to the original at the end of the function.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -30,73 +47,137 @@ dns_resolver2="8.8.4.4"
...
+	num_a=$(($1 << 24))
+	num_b=$(($2 << 16))
+	num_c=$(($3 << 8))
+	num_d=$4
 	IFS=$OLDIFS
+
+	num=$((num_a + $num_b + $num_c + $num_d))
+}
+
+num_to_ip()
```
bahamas10: can be cleaned up again to not leak variables by using arrays and `IFS` to preform the array.join emulation

``` bash
num_to_ip() {
        local num=$1
        local a=()

        # build ip addr (reversed)
        while true; do
                a+=("$((num & 255))")
                ((num >>= 8)) || break
        done

        # reverse array
        local i b=()
        for ((i = ${#a[@]} - 1; i >= 0; i--)) ; do
                b+=("${a[i]}")
        done

        # print the result
        IFS=.
        echo "${b[*]}"
        unset IFS
}

ip=$(num_to_ip 167772440)
# => "10.0.1.24"
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -141,52 +222,221 @@ is_net()
...
-	[ $d -ge 255 ] && return 1
+	return 0
+}
+
+# Tests if input is an email address
+is_email() {
+	regex="^[a-z0-9!#\$%&'*+/=?^_\`{|}~-]+(\.[a-z0-9!#$%&'*+/=?^_\`{|}~-]+)*@([a-z0-9]([a-z0-9-]*[a-z0-9])?\.?)+[a-z0-9]([a-z0-9-]*[a-z0-9])?\$"
+	ADDRESS=$1
+
+	[[ $ADDRESS =~ $regex ]] && return 0
```
bahamas10: the `return` statements are redundant here, whatever the return code of the last command ran in a bash function will be the return code of the entire function... 

``` bash
is_email() {
  local regex='...'
  local address=$1
  [[ $address =~ $regex ]]
}
```

rmustacc:
Whatever general clean up is happening against this likely should be synced back to sdc-headnode installer, where this was originally adapted from.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -141,52 +222,221 @@ is_net()
...
+		val=$(getanswer "${key}")
+		if [[ $? == 0 ]]; then
+			if [[ ${val} == "<default>" ]]; then
+				val=${def}
+			fi
+			return
+		fi
+	fi
+
+	if [ -z "$def" ]; then
```
bahamas10: personally, i'd prefer `[[` over `[` as it is a builtin and more robust than `[`... see http://mywiki.wooledge.org/BashFAQ/031 for more info

also, quotes around `$def` are not needed if `[[ ... ]]` is used.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -207,62 +457,105 @@ printnics()
...
+        return
+    fi
+
+    printnics
+    num=0
+    while [ /usr/bin/true ]; do
+        prmpt_str="Enter the number of the NIC for the $1 interface: "
+        printf "$prmpt_str"
+        read num
+        if ! [[ "$num" =~ ^[0-9]+$ ]] ; then
```
bahamas10: `isdigit` is defined in this file

``` bash
if ! isdigit "$num"; then
  ...
fi
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -271,55 +564,207 @@ promptpw()
...
-      printf "The disks %s are not valid choices" $bad
-    else
-      DISK_LIST="$val"
-      break
-    fi
-  done
+	states=(1)
+	#states[0]=1
+	while IFS=: read -r link state ; do
+		states=( ${states[@]-} $(echo "$state") )
```
bahamas10: you can append to an array with

``` bash
states+=("$state")
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -271,55 +564,207 @@ promptpw()
...
+    out += "total capacity:   " + Number(capacity / 1073741824).toFixed(2)
+      + " GB";
+    ' out
+}
 
+promptpool()
+{
+	local layout=""
+	local disks=
+	while [[ /usr/bin/true ]]; do
```
bahamas10: true is typically built into bash and it can be more efficient to call it directly

``` bash
while true; do
end
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -327,187 +772,158 @@ create_dump()
...
-  if ! echo $datasets | grep ${VARDS} > /dev/null; then
-    printf "%-56s" "Initializing var dataset... "
-    zfs create ${VARDS} || \
-      fatal "failed to create the var dataset"
-    chmod 755 /${VARDS}
-    cd /var
-    if ( ! find . -print | cpio -pdm /${VARDS} 2>/dev/null ); then
-        fatal "failed to initialize the var directory"
-    fi
+	datasets=$(zfs list -H -o name | xargs)
```
bahamas10: why `| xargs`... seems like relying on implicit behavior of xargs

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -327,187 +772,158 @@ create_dump()
...
-          # specified size is smaller than DRAM.$
-          #
-          size=${SYSINFO_MiB_of_Memory}
-          zfs create -V ${size}mb ${SWAPVOL}
-          swap -a /dev/zvol/dsk/${SWAPVOL}
-    fi
-    printf "%4s\n" "done"
-  fi
-}
+	if ! echo $datasets | grep dump > /dev/null; then
```
bahamas10: this could potentially match anything with the word dump in it... wouldn't you want something tighter like


``` bash
if ! grep -q '^zones/dump$' <<< "$datasets"; then
  # make dump zvole
fi
```

overlay/generic/smartdc/lib/smartos_prompt_config.sh
bahamas10: no, `IFS` is only set for the `read` invocation with its use here

konobi: ah yes... quite right

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -327,187 +772,158 @@ create_dump()
...
-  # Since there may be more than one storage pool on the system, put a
-  # file with a certain name in the actual "system" pool.
-  #
-  touch /${SYS_ZPOOL}/.system_pool
+	layout=$1
+	disks=$2
+	pool=zones
+
+	# If the pool already exists, don't create it again.
+	if /usr/sbin/zpool list -H -o name $pool >/dev/null 2>/dev/null; then
```
jdwoody: Could this be moved earlier in the process? I created a raidz2 on a large array prior to going through the initial smartos configuration. The current install process indicates that it's going to destroy the data even though it exits here.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -327,187 +772,158 @@ create_dump()
...
+		printf "%-56s\n" "Pool '$pool' exists, skipping creation... "
```
rmustacc: @jdwoody had the following comment, which I mistakenly deleted here:

```
Could this be moved earlier in the process? I created a raidz2 on a large array prior to going through the initial smartos configuration. The current install process indicates that it's going to destroy the data even though it exits here.
```

nshalman:
Rather than repeat myself over and over again on IRC, I'll just link here...
For people who want to test out this installer...

If you boot from a CD, grab and burn this ISO:
http://us-east.manta.joyent.com/nahamu/public/smartos/platform-hourly.iso

If you boot from a USB stick:
1. Grab this tarball: http://us-east.manta.joyent.com/nahamu/public/smartos/platform-hourly.tgz
2. Extract it onto your USB stick (it will be named "platform-```<datestamp>```")
3. Move aside the directory named "platform" to e.g. "platform-orig"
4. Rename "platform-```<datestamp>```" to "platform"
5. Boot from your USB stick

schmurfy:
@nshalman  Thanks for posting this here, I am very interested by this and will try to test that if I find a test server.

schmurfy:
do you have an example of what the answers file should look like ?

rmustacc:
@schmurfy We're not quite ready to support the answers.json in the new SmartOS installer quite yet, though this does give us the basic building blocks to do so.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
nshalman: @jdwoody could you clarify what the problem is with this newer installer?
I would have thought that you would select "manual", import the pool, then close the shell and everything would be fine.
Or did "current install process" refer to the current version being shipped rather than the current version of this patch?

schmurfy:
ok, thanks.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
jdwoody: This comment can be disregarded. "Current install process" did in fact refer to the current version being shipped. The changes in promptpool will alleviate this in the future. Thanks.

nshalman:
There is a ton of cleanup that could happen to this code that should also happen to the code in https://github.com/joyent/sdc-headnode/blob/master/scripts/prompt-config.sh
In theory the shared code could be refactored into a library of functions that both scripts could use, though then there's the new question of where that library should live.

Questions:
1. What is a good long-term solution for two very similar scripts that currently live in two different repos?
2. What is the minimal amount of remaining work for me (a non-Joyent employee) to get some version of this merged to stop the pain being reported on IRC?
3. Alternately, is there time and desire on the part of a Joyent employee to pick this back up and carry it across the finish line?

rmustacc:
For the moment, let's not worry about the long-term solution here and focus on getting this in, given the larger improvement here and the failure modes that people hit here. If we go over the outstanding review feedback, the primary question is how much of the feedback is correctness vs. style. @bahamas10 would you happen to have some time that you could clean this up?

If not, we should probably just go ahead and land this and then come back later, clean it up, and figure out the longer term strategy against the sdc-headnode copy in some regard.

@jclulow, @bahamas10  how does this sound to you?

bahamas10:
@rmustacc this sounds good to me.  @nshalman I'll go through this and do some cleanup based on the notes I have above and post a diff for consumption.

bahamas10:
@rmustacc @nshalman I've reviewed the code again and everything looks good in terms of correctness.  In terms of style there are some changes I would like to make (some in my comments, some elsewhere), but those changes apply to more than what is in this diff.  The bash written in this PR matches the style of the entire file so I'd be +1 on merging this now, and cleaning up the code(1) as a separate change in the future.

(1) wrt style only - no feature changes

nshalman:
Now that today's build has been built, should we try to land this per @bahamas10's feedback?

rmustacc:
Resolved in 93de422.

nshalman:
Heads up, @bixu and @sax! I don't know if this will break Vagrant stuff, but I figure I should flag it...

sax:
@nshalman thanks for the heads up! I don't *think* this will break existing vagrant boxes, but it will definitely effect the creation of new vagrant boxes. Our vagrant plugin can switch out the platform image to whatever is the latest, but my understanding is that these changes only effect new installations, not installations with a zpool already configured.

nshalman:
@sax correct, this shouldn't break existing vagrant boxes. It's probably worth opening an issue in some appropriate project to track this, but I wasn't sure which one...

sax:
I need to completely rework the project that I've been using to create boxes, so no worries about that.

PR #444, opened 2015-05-07
Docker Update

canvasio:


lloydde:
@canvasio looks like this pull request was made in error. What can I help you with?

canvasio:
Let me give it another go and if it doesn't work the 2nd time around!

lloydde:
@canvasio if you need a hand, hit us up on irc or email, https://github.com/joyent/sdc#community

canvasio:
I read a article on Wired this evening that peeked my curiosity. CoreOs decided to move away from Docker and implement their own container service (rkt). Is smartOS following suit by introducing Triton? Triton is a native container service. That can be used in conjunction with docker but isn't dependent on docker. So why not just offer Triton by itself and move away from docker. 


You might want to deflect  this question to Bryan Cantrill.

lloydde:
@canvasio I think you will find https://www.joyent.com/blog/dockers-killer-feature insightful. 

"The problem with this pair of assertions is that they neglect one essential aspect of Docker — one that is arguably its killer feature: its open API. In fact, when we at Joyent first took apart Docker, the robust API came as a pleasant surprise..."

That article is from Dec 02, 2014, but Bryan shared the same sentiments at last weeks SF Docker meetup, http://www.meetup.com/Docker-meetups/events/222051720/ .

PR #445, opened 2015-05-12
OS-5062 VM.load should use "zfs list -r" where possible

arekinath:
Currently, calling VM.load on one specific zone causes it to do a "zfs list" on all datasets and snapshots on the entire system. When we already know the zone is of the `joyent[-minimal]` or `lx` brands, we don't have to do this, as we can safely assume all the related datasets are under the zone's root dataset as children. So we can use `zfs list -r` and not have to walk the entire pool.

This makes `VM.load` (and thus `vmadm get` and also the SDC heartbeating) faster on systems with many ZFS datasets. It also reduces the CPU load generated by regular heartbeats and lock contention.

For example, on one of my CNs, before this patch:

```
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m1.679s
user	0m0.787s
sys	0m0.907s
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m1.655s
user	0m0.789s
sys	0m0.882s
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m1.680s
user	0m0.786s
sys	0m0.910s
```

And after this patch:

```
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m0.868s
user	0m0.456s
sys	0m0.409s
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m0.864s
user	0m0.456s
sys	0m0.406s
[root@a0-36-9f-29-7b-f4 (gps-1) ~]# time vmadm get $uuid >/dev/null
real	0m0.910s
user	0m0.475s
sys	0m0.414s
```

### DIFF
--- a/src/vm/node_modules/vmload/vmload-datasets.js
+++ b/src/vm/node_modules/vmload/vmload-datasets.js
@@ -178,7 +178,7 @@ function cleanDatasetObject(obj)
     }
 }
 
-function getDatasets(options, callback)
+function getDatasets(uuid, obj, options, callback)
 {
     var fields;
     var log;
@@ -261,6 +261,20 @@ function getDatasets(options, callback)
         log: log,
         types: zfs_types
     };
+    /*
+     * If this is about one specific zone, and it's a smartos/lx zone, we only
+     * need to do our zfs list recursively under the zone's dataset.
+     *
+     * This saves a lot of time on machines with lots of zfs datasets and snaps.
+     */
+    if (uuid && ['joyent', 'joyent-minimal', 'lx'].indexOf(obj.brand) !== -1) {
+        /*
+         * The zonepath property is always of the form '/' + zfs_filesystem,
+         * as this is how it is built in createZone().
+         */
+        var topzfs = obj.zonepath.slice(1);
+        _options.parent = topzfs;
+    }
     if (options.hasOwnProperty('spawnZfs')) {
         _options.spawnZfs = options.spawnZfs;
     }
@@ -288,6 +302,9 @@ function getZfsList(options, callback) {
     if (options.hasOwnProperty('spawnZfs')) {
         task.spawnZfs = options.spawnZfs;
     }
+    if (options.hasOwnProperty('parent')) {
+        task.parent = options.parent;
+    }
 
     try {
         zfs_list_in_progress[task].on('result', callback);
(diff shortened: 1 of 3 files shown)


src/vm/node_modules/vmload/vmload-datasets.js
```diff
@@ -337,7 +350,11 @@ function zfs(options, cmd, args, lineHandler, callback)
 
     // we don't expect data on stderr, so treat as a warning
     zfs_child.stderr.on('data', function (data) {
-        log.warn({stderr: data.toString()}, 'zfs[' + zfs_child.pid + '] wrote '
+        var str = data.toString();
+        if (str.indexOf('dataset does not exist') !== -1) {
```
jclulow: If you're going to interpret the `stderr` output of `zfs(1M)`, we should sanitise the environment and force the C locale.  For example: [tools/lib/oscmds.js](https://github.com/joyent/sdc-headnode/blob/master/tools/lib/oscmds.js#L28-L57) in `sdc-headnode`.

Interpreting the value passed to a single "`data`" event is also not the right way to consume a stream.  You have to assume that the string you're looking for may straddle multiple events.  You should accumulate the data until a line termination, or the "`end`" event of the stream.

arekinath:
Amended with @jclulow's suggestions. I just copied that function in from sdc-headnode -- we want exactly the same thing, so...

src/vm/node_modules/vmload/vmload-datasets.js
```diff
@@ -261,6 +261,15 @@ function getDatasets(options, callback)
...
         types: zfs_types
     };
+    /*
+     * If this is about one specific zone, and it's a smartos/lx zone, we only need
+     * to do our zfs list recursively under the zone's dataset.
+     *
+     * This saves a lot of time on machines with lots of zfs datasets and snaps
+     */
+    if (uuid && (brand === 'joyent' || brand === 'joyent-minimal' || brand === 'lx')) {
+        _options.parent = 'zones/' + uuid;
```
bahamas10: I believe it is possible to have SmartOS use a different root dataset than "zones"... this may break for those users.

arekinath:
Amended to use `zonepath` rather than assuming zones are all on the `zones` pool

src/vm/node_modules/vmload/vmload-datasets.js
```diff
@@ -261,6 +261,20 @@ function getDatasets(options, callback)
...
+     * If this is about one specific zone, and it's a smartos/lx zone, we only
+     * need to do our zfs list recursively under the zone's dataset.
+     *
+     * This saves a lot of time on machines with lots of zfs datasets and snaps.
+     */
+    if (uuid && ['joyent', 'joyent-minimal', 'lx'].indexOf(obj.brand) !== -1) {
+        /*
+         * The zonepath property is always of the form '/' + zfs_filesystem,
+         * as this is how it is built in createZone().
+         */
```
bahamas10: perhaps add an assert here to compliment the comment

``` js
assert.equal(obj.zonepath[0], '/', 'invalid obj.zonepath');
```

src/vm/node_modules/vmload/vmload-datasets.js
```diff
@@ -337,8 +387,17 @@ function zfs(options, cmd, args, lineHandler, callback)
 
     // we don't expect data on stderr, so treat as a warning
     zfs_child.stderr.on('data', function (data) {
```
bahamas10: it's possible (not likely) to get incomplete data here, and possibly break on only half a line... it may be worth it to store all of the stderr then post-process it in the exit handler.

bahamas10:
I have not tested this, but AFAICT the code itself looks good to me - intent is clear and there are no obvious problems.

arekinath:
Merged as 62c472bf6ee6c2d2ff8eb34f16c78ff2b7b136f3

PR #474, opened 2015-08-13
Fix typo in vmadm manual

michaelnisi:
After stumbling over this at least three times today, I might as well fix it.


bahamas10:
Thanks!

Tracked in https://smartos.org/bugview/OS-4707
Fixed in https://github.com/joyent/smartos-live/commit/c3881f2255b727b0b07d6a33a6f38bc327c57ccf

Issue #464, opened 2015-07-24
support for ipv6 on admin_nic


sjorge:
In addition to #231 support in the gz for the admin and external nics to have ipv6 addresses and gateway+ dns would be awesome.

melloc:
Resolved by 0494fec. Work done by @sjorge in #475.

Issue #431, opened 2015-03-30
No static IPv6 in LX zones


calmh:
Using autoconfiguration works (with the appropriate filtering turned off), by starting `/native/usr/lib/inet/in.ndpd`; however manual configuration seems impossible.

```
root@dl2:~# /native/sbin/ifconfig 
lo0: flags=2001000849<UP,LOOPBACK,RUNNING,MULTICAST,IPv4,VIRTUAL> mtu 8232 index 1
        inet 127.0.0.1 netmask ff000000 
eth0: flags=1000843<UP,BROADCAST,RUNNING,MULTICAST,IPv4> mtu 1500 index 2
        inet 172.16.32.21 netmask ffffff00 broadcast 172.16.32.255
        ether 72:c4:cc:e5:b3:12 
lo0: flags=2002000849<UP,LOOPBACK,RUNNING,MULTICAST,IPv6,VIRTUAL> mtu 8252 index 1
        inet6 ::1/128 
eth0: flags=2000841<UP,RUNNING,MULTICAST,IPv6> mtu 1500 index 2
        inet6 fe80::70c4:ccff:fee5:b312/10 
        ether 72:c4:cc:e5:b3:12 
eth0:1: flags=2080841<UP,RUNNING,MULTICAST,ADDRCONF,IPv6> mtu 1500 index 2
        inet6 2001:470:deeb:32:70c4:ccff:fee5:b312/64 
root@dl2:~# ifconfig eth0 add 2001:470:deeb:32::21/64
SIOCSIFADDR: No such device or address
root@dl2:~# ip addr add 2001:470:deeb:32::21/64 dev eth0
Cannot find device "eth0"
root@dl2:~# /native/sbin/ifconfig eth0 inet6 addif 2001:470:deeb:32::21/64
ifconfig: 2001:470:deeb:32::21/64: bad address
root@dl2:~# /native/sbin/ipadm create-addr -t -T static -a 2001:470:deeb:32::21/64 eth0/v6s
ipadm: Error in setting local address: Operation failed
root@dl2:~# 
```

joyent_20150320T034246Z, lx-ubuntu-14.04 20150320

pfmooney:
Using the built-in linux (ifconfig/ip) utilities for network configuration is not yet supported.
The native ifconfig will work, but it must be chrooted to access certain assets at their expected paths.
The /sbin/ifconfig-native script is pushed into the zone for that purpose.

Let me know if that doesn't work for you.

[...]

PR #475, opened 2015-08-15
Globalzone IPv6 support

sjorge:
## Globalzone IPv6 support
This should enable IPv6 support for the admin and external NICs. Fixing #464 

Paging @wiedi for testing.
Platform image available here: http://sjorge.sinners.be/platform-20150814T174741Z.iso 

#### pure autoconfiguration
```
admin_ip6=addrconf
```
#### static (also does auto configuration as in.ndpd needs to be running)
```
admin_ip6=2001:6f8:1480:10::123
```
Overwrite the default prefix (/64)
```
admin_ip6=2001:6f8:1480:10::123/48
```

### notes
```admin_gateway6``` is also supported, however in.ndpd will overwrite the default route if one is present. 

I have found no way to prevent this. (route flush -inet6 has no effect as in.ndpd runs **after** network/physical) There is no equivalent of /etc/defaultrouter6.

This is implemented to use in a pure static environment without a router advertising prefixes.

### DIFF
--- a/overlay/generic/lib/svc/method/net-physical
+++ b/overlay/generic/lib/svc/method/net-physical
@@ -405,6 +405,16 @@ if smf_is_globalzone; then
         admin_netmask=${BOOT_admin_netmask}
     fi
 
+    admin_ip6=${CONFIG_admin_ip6}
+    if [[ -z "$admin_ip6" ]]; then
+        admin_ip6=${BOOT_admin_ip6}
+    fi
+
+    admin_netmask6=${CONFIG_admin_netmask6}
+    if [[ -z "$admin_netmask6" ]]; then
+        admin_netmask6=${BOOT_admin_netmask6}
+    fi
+
     if [[ -n ${admin_ip} ]] && [[ -n ${admin_netmask} ]];
     then
         /sbin/ifconfig ${SYSINFO_NIC_admin} inet ${admin_ip} \
@@ -444,6 +454,31 @@ if smf_is_globalzone; then
 
             ADMIN_NIC_UP=true
         fi
+    fi
+
+    if [[ -n ${admin_ip6} ]]; then
+        # NOTE: plumbing the interface for IPv6 also implicitly sets a
+        # default gateway, this is due to in.ndpd and can't be disabled
+        if [[ -n ${admin_ip6} ]] && [[ -n ${admin_netmask6} ]]; then
+            # Plumb interface for inet6
+            ifconfig ${SYSINFO_NIC_admin} inet6 plumb up
+
+            # Configure static IPv6
+            /sbin/ifconfig ${SYSINFO_NIC_admin} inet6 \
+                addif ${admin_ip6}/${admin_netmask6} up
+            ADMIN_NIC_UP=true
+        else
+            if [[ ${admin_ip6} == "auto" ]]; then
+                # Plumb interface for inet6
+                ifconfig ${SYSINFO_NIC_admin} inet6 plumb up
+
+                # Autodiscovery IPv6 using SLAAC
+                # in.ndpd will be started later, we just need to plumb the interface
+                ADMIN_NIC_UP=true
+            fi
+        fi
+
+        # don't setup resolv.conf, IPv6 addresses already work
     fi
 
     # If on Parallels or VirtualBox, create a bridge which
--- a/overlay/generic/lib/svc/method/net-physical
+++ b/overlay/generic/lib/svc/method/net-physical
@@ -219,6 +219,7 @@ function vnic_up
     fi
     eval "vnic_${iface}_up=true"
 }
+
 function vnic_up6
 {
     set -o xtrace
--- a/overlay/generic/lib/svc/method/net-physical
+++ b/overlay/generic/lib/svc/method/net-physical
@@ -226,16 +226,15 @@ function vnic_up6
     typeset link="$1"
     typeset iface="$2"
     typeset ip="$3"
-    typeset netmask="$4"
-    typeset vlan_id="$5"
-    typeset mac_addr="$6"
-    typeset mtu="$7"
-    typeset details="link='${link}', iface='${iface}', ip='${ip}', netmask='${netmask}, vlan_id='${vlan_id}'"
+    typeset vlan_id="$4"
+    typeset mac_addr="$5"
+    typeset mtu="$6"
+    typeset details="link='${link}', iface='${iface}', ip6='${ip}', vlan_id='${vlan_id}'"
     typeset vlan_opt=
     typeset mac_addr_opt=
     typeset prop_opt=
 
-    if [[ -z ${link} ]] || [[ -z ${iface} ]] || [[ -z ${ip} ]] || ([[ ${ip} != "addrconf" ]] && [[ -z ${netmask} ]]); then
+    if [[ -z ${link} ]] || [[ -z ${iface} ]] || [[ -z ${ip} ]]; then
         echo "WARNING: not bringing up nic (insufficient configuration): $details"
         return
     fi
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: overlay/generic/lib/svc/method/net-physical [hunk 1]
```diff
@@ -219,6 +219,74 @@ function vnic_up
...
+        if [[ -n ${mac_addr} ]]; then
+            mac_addr_opt="-m ${mac_addr}"
+        fi
+
+        if [[ -n ${mtu} ]]; then
+            valid_mtu ${iface} ${mtu}
+            prop_opt="-p mtu=${mtu}"
+        fi
+
+        /usr/sbin/dladm create-vnic -t -l ${link} ${prop_opt} ${vlan_opt} \
```
bahamat: Why create a vnic? It should use the same nic as the existing v4 interface.

sjorge: Because you external does can work with just IPv6.

We reuse the one created by vnic_up, but we create it if there is none.

E.g. Only setting ```external_ip6``` works and has been tested.

bahamat: Oh, ok. I see what you're doing now.

bahamat:
I'll hot patch my HN and CN in my test rig and try this out as soon as I can.

overlay/generic/lib/svc/method/net-physical
```diff
@@ -219,6 +219,74 @@ function vnic_up
...
+        fi
+    fi
+
+    /sbin/ifconfig ${iface} inet6 plumb
+    if [[ $? -ne 0 ]]; then
+        echo "Failed to plumb ${iface}"
+        exit $SMF_EXIT_ERR_FATAL
+    fi
+
+    if [[ ${ip} == "auto" ]] || [[ -n ${ip} ]] && [[ -n ${netmask} ]]; then
```
bahamas10: can you explain the intent with this?  Bash will evaluate the || expression before && because of its order of operations (left to right in this case), so it currently reads like "if ip is set to "auto" or ip is set to anything AND netmask in set to anything, do ..." - is that a correct assessment?

If it is, you can remove the check of $ip being set and reduce it all to

``` bash
if [[ $ip == 'auto' && -n $netmask ]]; then
    ....
fi
```

If not, it may be clearer to join some of the `[[...]]` to create logical separations, ie.

``` bash
if [[ -n $ip && -n $netmask ]] || [[ $ip == 'auto' ]]; then
    ...
fi
```

That means if both ip and netmask are set, or if ip is set to auto, do `...`

bahamas10: edit: i see this script is `ksh` actually, but the order of operations is the same for both shells

sjorge: The plumbing should happen when ip == auto OR when ip AND netmask are set.

I'll update it so it is more readable.

sjorge:
@bahamas10 more commits landed, iso for testing also update.

wiedi:
Thanks a lot for doing this, this is great!

I have not looked too closely at the source yet but it is on my todo list. Just started some basic testing and wanted to report some feedback.

- `admin_ip6=auto` works with SLAAC
- `external_ip6=<static>` also works

Regarding the `_netmask6` name: in IPv6 the term usually used is "prefix length" or "prefixlen" instead of netmask. As long as no one is using it yet it might be easier to change now.
It might also possible to drop this option line completly by using the `<ip>/<prefixlen>` syntax for the `_ip6` field. If no prefixlen is specified we could default to 64.
Not sure what others think? (I don't care that much and don't want to start a bikeshed, just wanted to mention it early)

What I could not get to work so far (not sure if it is even supposed to work yet) is IPv6 addresses on any other nictag (etherstub or physical). Maybe I'm not using the correct config syntax.

I also have not tested DHCPv6 and the gateway option yet, but will do so once I have more test hardware configured.

Thanks again!

sjorge:
I first had it as _prefix but though _netmask6 (although technically not the correct term) fitted better overall.

When testing the gateway make sure to read the little caveat above, it get's set but when SLAAC is working on the network, the one it gets via SLAAC will always win for some reason. 

I never had any luck with other nictags than external/admin... with both ipv4 and ipv6... Since I had to implement vnic_up6 I also made the changes to do to the other tags but never got anything on them, not even v4 either.

bahamat:
I would highly prefer protocol specific terms rather than just what people are used to. "Prefix" or "prefix length" is preferred over "subnet" and "addrconf" (to match ipadm) is preferred over "dhcp" or "slaac".

See #431 for work being done to add IPv6 to zones. GZ and non-GZ terms should match.

sjorge:
I will drop _netmask6, and rename the auto value to addrconf. Will try to do those tomorrow if time permits

sjorge:
@wiedi seems /64 is the default for ipadm and ifconfig if the prefix is missing, this simplifies things a lot. 2nd commit coming some after testing

sjorge:
All done, uploading testing iso now.

overlay/generic/lib/svc/method/net-physical
```diff
@@ -522,6 +647,34 @@ if smf_is_globalzone; then
                 fi
             done
 
+            for key in ${config_ip6_keys}; do
+                if [[ ${key} == ${tag}[0-9]_ip6 ]] || [[ ${key} == ${tag}[0-9][0-9]_ip6 ]]; then
+                    iface=${key//_i6p/}
```
wiedi: small typo, this should be `iface=${key//_ip6/}`

sjorge:
Nice catch @wiedi , uploading new iso now

wiedi:
I now have tested with success the values "addrconf", ip without prefixlen, ip/64 and ip/48 for admin, external and other physical nictags.

If there are more ideas for things I could try please let me know, otherwise I'm happy to see this merged.

sjorge:
Thanks for verifying! Let's hope this get merged :3

rmustacc:
@sjorge between @melloc and myself, we'll make sure to take a look at this in the coming week.

sjorge:
Sounds good, I will be traveling with limited internet so if you need something from me, my replies will be slow.

sjorge:
@rmustacc poke, would be nice to see this appear soon too so there is full ipv6 support. Now that the zone bits have landed.

rmustacc:
@sjorge You'll want @melloc to review this.

sjorge:
Well then @melloc, poke :3

overlay/generic/lib/svc/method/net-physical
```diff
@@ -219,6 +219,73 @@ function vnic_up
     fi
     eval "vnic_${iface}_up=true"
 }
```
melloc: @sjorge Could you add a blank line here, as well as a comment above `vnic_up6`?

overlay/generic/lib/svc/method/net-physical
```diff
@@ -219,6 +219,73 @@ function vnic_up
...
+{
+    set -o xtrace
+
+    typeset link="$1"
+    typeset iface="$2"
+    typeset ip="$3"
+    typeset vlan_id="$4"
+    typeset mac_addr="$5"
+    typeset mtu="$6"
+    typeset details="link='${link}', iface='${iface}', ip6='${ip}', vlan_id='${vlan_id}'"
```
melloc: @sjorge This line, along with several others, are greater than 80 characters. It looks like there are already some long lines in this file, but I'd like to avoid adding more. Can you break the new lines up?

melloc:
@sjorge I think that this looks good. I'm going to download the changes locally to play around with it. What type of tests have you done, invalid/garbage inputs tested, etc.?

sjorge:
@melloc fixed up the stuff you asked for, doing a new build now but it won't finish today (local time)
I will do my usual tests tomorrow after work. Those are:
 * test without ip6 commands
 * test with ip6 set to addrconf
    * test reachable from host in same subnet
    * test if host in same subnet is reachable
    * test if host in different subnet is reachable
  * test with ip6 set to a static addr
   * see tests as above
  * test with bogus ip6 set

melloc:
@sjorge Sounds good! Thanks for doing this! Once you've finished your tests tomorrow, let me know.

sjorge:
My build failed, will debug once I get home from work

sjorge:
My tests look good, same result as before the changes you asked for.

sjorge:
/me pokes @melloc 
Would love to stop using custom builds ;)

melloc:
Hey @sjorge ! I've incorporated your changes into a local commit and done several tests. There are several more I want to do before pushing, but they'll be in master soon.

sjorge:
Thanks for the update :) wasn't sure if you had actually started doing some local tests or not.

melloc:
Put back into master by 0494fec.

Issue #484, opened 2015-09-25
installer creates invalid dump zvol, prevents system boot


failedrequest:
The current code for smartos_prompt_config.sh in the function create_dump fails to set the dump zvol to a size suitable for dumpadm to work with.  As a result svc:/system/filesystem/smartdc:default fails to start yells about this.

 svc:/system/filesystem/smartdc:default (Joyent file system mounts)
  See: /etc/svc/volatile/system-filesystem-smartdc:default.log
 + zpool set feature@multi_vdev_crash_dump=enabled zones
+ dumpadm -y -d /dev/zvol/dsk/zones/dump
+ to small to hold a system dumpdump size 5151584256 bytes, device size 4294967296 bytes
 + fatal 'failed to configure dump device'
+ echo 'Error: failed to configure dump device'
Error: failed to configure dump device

After some testing it appears that setting the dump zvol to a 14G volsize or larger makes the check work and the system complete a bootup. 

This was noted on platfom images 20150423T155306Z and newer .

nshalman:
In conversation on IRC @rmustacc's response to my suggestion of using the code from [joysetup.sh#L361-375](https://github.com/joyent/sdc-headnode/blob/89e7d2dc747a68e85d9f2aee5202cada44cdae03/scripts/joysetup.sh#L361-L375),  was: "Probably something like that, though constraining it based on some amount of available disk as well for smartos."

failedrequest:
I would say rather then waste the disk space make the cap 32G for now or 33% the ram .

failedrequest:
All
 I reviewed the issue on a fresh install. This is what the installer errors out with on the latest platform image 20150929T222233Z


<pre>
Creating pool zones...                                  done
Making dump zvol...                                     dumpadm: dump device /dev/zvol/dsk/zones/dump is too small to hold a system dump
dump size 5715423232 bytes, device size 4294967296 bytes
</pre>

That warning is from dumpadm , If the installer were to run dumpadm -e to get a dump volsize and then set it according to a multiple of the zfs blocksize then we end up with a working setup. 

So in the above noted case setting the volsize to 6G , ( Rounding 5.38G  up to 5.53G here ) works . Also it should be noted that dumpadm is setup to only collect kernel pages by default so we don't need a large dump vol . If someone were to change dumpadm to all , then it also up to them to set the dump volsize to 50% of the ram to hold the full dump.


Here is my proposed change 

<pre>
create_dump()
{

[...]

PR #488, opened 2015-10-01
joyent/smartos-live#484 Fresh install fails to boot completely on large ram servers

failedrequest:
This is a small patch to the fix Fresh install fails to boot completely due to bug in code for setting dump zvol  issue #484  

In this patch we use dumpadm -e to get the estimated dumpsize for kernel pages. Since SmartOS is configured to only dump kernel pages by default this is a safe size to use.  The value from dumpadm -e is converted to MB and padded by 1mb to allow it to fix into the proper volblock . We also check to insure that the minimum size for the dumpvol is 4G . 

.

### DIFF
--- a/overlay/generic/smartdc/lib/smartos_prompt_config.sh
+++ b/overlay/generic/smartdc/lib/smartos_prompt_config.sh
@@ -751,18 +751,23 @@ promptpool()
 
 create_dump()
 {
-	# Get avail zpool size - this assumes we're not using any space yet.
-	base_size=`zfs get -H -p -o value available ${SYS_ZPOOL}`
-	# Convert to MB
-	base_size=`expr $base_size / 1000000`
-	# Calculate 5% of that
-	base_size=`expr $base_size / 20`
-	# Cap it at 4GB
-	[ ${base_size} -gt 4096 ] && base_size=4096
+	# The original setup capped the dumpvol at 4G however this is too small on
+        # large memory servers.  This function will setup a dumpvol based on the 
+        # size of the dumpadm -e estimate.  We take value and double it to insure
+        # we have a good sized dumpvol, out of the box for most cases. We also set
+        # the smallest dumpvol 1G. The reason to double the estimate is this is
+        # the cleanest way to get a usable dumpsize size on most systems. Setting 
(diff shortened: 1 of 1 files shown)


nshalman:
Also usually a good idea to update the commit message to start with the bug ID being fixed. In this case it should be:
```
joyent/smartos-live#484 Fresh install fails to boot completly due to bug in code for setting dump zvol
```

If you dislike that summary you might want to edit the description of the original bug first, then set the commit to match.

failedrequest:
Updated the code to use 2 times the estimate from dumpadm -e . It has been tested on 256G ram servers and it resolves the issue as expcted.

overlay/generic/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -751,18 +751,16 @@ promptpool()
...
-	[ ${base_size} -gt 4096 ] && base_size=4096
-
-	# Create the dump zvol
-	zfs create -V ${base_size}mb ${SYS_ZPOOL}/dump || \
-	    fatal "failed to create the dump zvol"
+	# Get the estimated dump size from dumpadm -e in mb double it and convert to int
+
+	local dumpsize=$(echo "(( $(dumpadm -e |  sed "s/[^0-9|.]//g") * 1024) * 2 )/1" |bc)
+
+	# Ensure the dumpvol is at least 4G. This will store most kernel page only dumps
```
rmustacc: Where does 4 GiB come from?

failedrequest: The 4G was changed to 1G after testing showed that 4G was a waste. The current version uses 1G as the smallest size. Which should be sufficient for a starting point for dumps.

jclulow:
Thanks for the changes.  I will pull this into `master`.

jclulow:
Note that in final testing I found one last issue -- the dump size estimation is formatted by `zfs_nicenum()`, which will potentially render the size in megabytes (with a suffix of `M`), or even kilobytes (`K`).  I have adjusted the estimation calculation to take this into effect, and reformatted the comment a little.

Thanks again!

RSully:
This was done in commit 665689294042901fa6514919b70d2234a6a5eab4. This pull can probably be closed.

PR #496, opened 2015-10-19
Fix typo

erikengineer:


bahamas10:
thanks!

merged https://github.com/joyent/smartos-live/commit/c82997c6bb85a324ae682b6eb95d0d84ff1d06e2

internal bug: https://smartos.org/bugview/OS-4893

PR #523, opened 2015-12-01
OS-4990 SMF import example should use lib

### LINKED ISSUE #522: joyent-minimal brand documentation example import uses wrong location.
The example has:

> For example, if you were going to import the cron service you would run:
> 
> svccfg import /var/svc/manifest/system/cron.xml

Which should be:

> For example, if you were going to import the cron service you would run:
> 
> svccfg import /lib/svc/manifest/system/cron.xml

bahamat:
Fixes #522


bahamas10:
https://github.com/joyent/smartos-live/commit/76e8d500fa3c7e0bc31f065f980f602720517841

PR #529, opened 2015-12-18
The packages not from pkgsrc are not signed

nshalman:
The build instructions wiki page (https://wiki.smartos.org/display/DOC/Building+SmartOS+on+SmartOS) directs people to use base-multiarch-lts 14.4.X.
The pkgsrc trees in from 2014Q4 and later use signed packages and the pkgsrc configuration is set to verify signatures.

Unfortunately part of the configure step for building SmartOS pulls down some unsigned packages that are needed to set up the build environment.

This causes the configure step to be interactive rather than fully automatic.

People were getting confused on IRC and @jperkin suggested a simple workaround that is included in this pull request.

I believe this is important because I'm pretty sure that mountain-gorilla expects the configure step to run to completion on its own rather than be interactive.

As my additional comment below indicates, I updated the wiki page so that people will stop being surprised, but I think that we should fix this.

### DIFF
--- a/configure
+++ b/configure
@@ -109,7 +109,8 @@ function install_pkgin
 			curl -k "$conf_pkgsrcurl/$pkg.tgz" -o \
 			    /var/tmp/$pkg.tgz || fatal \
 			    "failed to fetch $pkg.tgz"
-			$conf_priv pkg_add /var/tmp/$pkg.tgz || fatal \
+			# The "-C /dev/null" is because these packages are not signed
+			$conf_priv pkg_add -C /dev/null /var/tmp/$pkg.tgz || fatal \
 			    "failed to pkg_add $pkg"
 			rm -f /var/tmp/$pkg.tgz
 		fi


nshalman:
A workaround is documented on the [wiki page](https://wiki.smartos.org/display/DOC/Building+SmartOS+on+SmartOS) (see [relevant change](https://wiki.smartos.org/pages/diffpages.action?pageId=754234&originalId=4948019))

That note can/should be removed after this issue is resolved.

nshalman:
Ping @trentm. I think this will be important if/when you move the build system to use zones using a pkgsrc tree with signed packages.

trentm:
(irc discussion: http://echelog.com/logs/browse/smartos/1452639600)

jperkin:
FWIW here's my position on this:

* The current state means that the user is forced to perform manual steps to get past this point when using 14.4.x onwards, due to those releases requiring packages be signed by default.  This is sub-optimal, but not critical.

* I am happy to sign the packages, but doing so will mean that anyone building the platform on pre-14.4.x will run into problems, as the pkg_add in those releases does not support the signed package.  They would need to manually install from an archived package or similar.  My guess is that this would be untenable at this stage, and I have no insight into how many installs would be affected by this.

* The workaround proposed by @nshalman disables the requirement that the packages be signed, by disabling the `VERIFIED_INSTALLATION` setting.  As this is a one-time operation against known packages, this should be a relatively safe operation, and does not affect the setting of `VERIFIED_INSTALLATION` for other instances of pkg_add/pkgin.  This workaround is suitable for both pre- and post- 14.4.x installs.  I'd be happy to see this patch go in, but am not involved in platform builds so it's not my call.

* At a later point when we can guarantee that everyone and everything building the platform is on at least 14.4.x, we can sign the packages, remove the `-C /dev/null` workaround, and have everything Just Work.

nshalman:
Resolved in https://github.com/joyent/smartos-live/commit/13f5ba1 (See also https://smartos.org/bugview/OS-5651)

PR #542, opened 2016-01-13
joyent/smartos#541 USB image partition table is not 4k aligned

nshalman:
Tests confirm that this both still boots and in some cases improves performance.


nshalman:
Don't merge this yet. Some people are still seeing oddities.

nshalman:
Apparently there was some confusion and this was found by all testers not to improve performance after all.
Closing unmerged.

PR #543, opened 2016-01-14
OS-4740 Docker containers up more than 10 seconds should have their d…

joshwilsdon:
…elay reset

This adds a new internal_metadata property 'docker:restartdelay' which gets set when *starting* a zone and will be used for the *next* start.

When vmadmd is going to restart a zone, it loads the restartdelay if there is one. It also loads the lastbooted and lastexited files which allow us to determine how long the VM was running. If the VM was running for longer than 10 seconds, we use the minimum delay (100ms) and set the docker:restartdelay to 200 (for the next restart). If the VM was running less than 10 seconds, or we cannot determine the runtime, we first look for the restartdelay. If that's not set, we use `100 * Math.pow(2, restartcount)` which which matches what docker does.

This commit also adds a test which exercises this functionality. It boots a docker zone with a cmd that exits some number of times immediately so we're under the 10 second limit and the delay should be increasing. Then it does a sleep 15 and exits and starts this pattern over. This way we can ensure:

 * the restartcount increases each time
 * the restartdelay increases until it's up 15 seconds, then is reset
 * the restartdelay increases again after reset
 * the VM is actually being restarted and the delays are actually (mostly) increasing

The actual boot delay is pretty variable because of all the steps involved when booting the zone. As such, the small delays are often hidden in the noise. That's why we allow it to double 7 times, to ensure that it's actually growing.

### DIFF
--- a/src/vm/sbin/vmadmd.js
+++ b/src/vm/sbin/vmadmd.js
@@ -648,7 +648,7 @@ function restartDockerContainer(uuid, opts)
 {
     var restart_delay;
 
-    assert(typeof(opts) === 'object', 'opts must be object');
+    assert(typeof (opts) === 'object', 'opts must be object');
 
     restart_delay = opts.delay;
 
--- a/src/vm/sbin/vmadmd.js
+++ b/src/vm/sbin/vmadmd.js
@@ -47,7 +47,18 @@ var url = require('url');
 var util = require('util');
 var vasync = require('vasync');
 
-var DOCKER_RUNTIME_DELAY_RESET = 10000; // ms zone being up, we zero delay
+/*
+ * The DOCKER_RUNTIME_DELAY_RESET parameter is used when restarting a Docker VM
+ * according to its restart policy. As with Docker, we have an increasing delay
+ * between each restart attempt for a container. The exception to this
+ * increasing delay is when a container stays running for some amount of time.
+ * The DOCKER_RUNTIME_DELAY_RESET is what controls this. If a docker VM was
+ * running for longer than this many milliseconds when it stopped, we'll reset
+ * the delay time between restarts. On Docker this is always 10 seconds, so we
+ * use that here too.
+ */
+var DOCKER_RUNTIME_DELAY_RESET = 10000;
+
 var UPGRADE_SCRIPT = '/smartdc/vm-upgrade/001_upgrade';
 var VMADMD_PORT = 8080;
 var VMADMD_AUTOBOOT_FILE = '/tmp/.autoboot_vmadmd';
@@ -133,15 +144,16 @@ function addLastRuntime(vmobj, opts, callback)
     var lastexited_mtime;
     var zonepath = vmobj.zonepath;
 
+    assert(vmobj, 'missing vmobj');
     assert(log, 'missing log'); // log is GLOBAL!
     assert(zonepath, 'missing zonepath');
 
-    lastbooted_filename = path.join(vmobj.zonepath, '/lastbooted');
-    lastexited_filename = path.join(vmobj.zonepath, '/lastexited');
+    lastbooted_filename = path.join(vmobj.zonepath, 'lastbooted');
+    lastexited_filename = path.join(vmobj.zonepath, 'lastexited');
 
     vasync.pipeline({funcs: [
-        function (_, cb) {
-            fs.stat(lastbooted_filename, function (err, st) {
+        function _statLastbooted(_, cb) {
+            fs.stat(lastbooted_filename, function _statLastbootedCb(err, st) {
                 if (err) {
                     log.error({err: err, vm_uuid: vmobj.uuid},
                         'failed to stat lastbooted');
@@ -151,8 +163,8 @@ function addLastRuntime(vmobj, opts, callback)
                 lastbooted_mtime = st.mtime.getTime();
                 cb();
             });
-        }, function (_, cb) {
-            fs.stat(lastexited_filename, function (err, st) {
+        }, function _statLastexited(_, cb) {
+            fs.stat(lastexited_filename, function _statLastexitedCb(err, st) {
                 if (err) {
                     log.error({err: err, vm_uuid: vmobj.uuid},
                         'failed to stat lastexited');
@@ -163,7 +175,7 @@ function addLastRuntime(vmobj, opts, callback)
                 cb();
             });
         }
-    ]}, function (err) {
+    ]}, function _addLastRuntimePipelineCb(err) {
         if (!err) {
             assert((lastexited_mtime - lastbooted_mtime) > 0,
                 'mtime delta must be positive [' + lastexited_mtime + ' <= '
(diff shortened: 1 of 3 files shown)


src/vm/sbin/vmadmd.js
```diff
@@ -45,7 +45,9 @@ var qs = require('querystring');
 var SyseventStream = require('/usr/vm/node_modules/sysevent-stream');
 var url = require('url');
 var util = require('util');
+var vasync = require('vasync');
 
+var DOCKER_RUNTIME_DELAY_RESET = 10000; // ms zone being up, we zero delay
```
misterdjules: The comment is not very clear without reading the PR's description, maybe writing a longer comment similar to what's described in the PR's description would help.

src/vm/sbin/vmadmd.js
```diff
@@ -116,6 +118,66 @@ function zonecfg(args, callback)
...
+ */
+function addLastRuntime(vmobj, opts, callback)
+{
+    var lastbooted_filename;
+    var lastbooted_mtime;
+    var lastexited_filename;
+    var lastexited_mtime;
+    var zonepath = vmobj.zonepath;
+
+    assert(log, 'missing log'); // log is GLOBAL!
```
misterdjules: Should we asset on `vmobj` too?

src/vm/sbin/vmadmd.js
```diff
@@ -116,6 +118,66 @@ function zonecfg(args, callback)
...
+    assert(zonepath, 'missing zonepath');
+
+    lastbooted_filename = path.join(vmobj.zonepath, '/lastbooted');
+    lastexited_filename = path.join(vmobj.zonepath, '/lastexited');
```
misterdjules: The leading `/` for the last path components should not be used, as `path.join` already does that formatting, and it would defeat the purpose of using the cross-platform `path.join`.

src/vm/sbin/vmadmd.js
```diff
@@ -116,6 +118,66 @@ function zonecfg(args, callback)
...
+
+    vasync.pipeline({funcs: [
+        function (_, cb) {
```
misterdjules: Naming these functions might help debugging.

src/vm/sbin/vmadmd.js
```diff
@@ -552,8 +614,10 @@ function rotateKVMLog(vm_uuid)
     }, 30 * 1000);
 }
 
-function restartDockerContainer(uuid, restart_count)
+function restartDockerContainer(uuid, opts)
 {
+    var restart_delay = opts.delay;
```
misterdjules: Would asserting on the type of `opts` help here, since the type changed?

src/vm/sbin/vmadmd.js
```diff
@@ -775,7 +853,8 @@ function updateZoneStatus(ev)
             'internal_metadata',
             'state',
             'uuid',
-            'zone_state'
+            'zone_state',
+            'zonepath'
         ]}, function (err, vmobj) {
             log.info(ev.zonename + ' (docker) went from ' + ev.oldstate + ' to '
```
misterdjules: Not necessarily related to this PR, but why is `err` not handled?

src/vm/tests/test-docker.js
```diff
@@ -1121,3 +1123,194 @@ test('test docker VM with good zlog_max_size', function (t) {
...
+    var cycle_reset_delay = 15; // seconds, should be >= 10
+    var emitter = new EventEmitter();
+    var events = [];
+    var payload = JSON.parse(JSON.stringify(common_payload));
+    var se;
+    var state = {brand: payload.brand};
+    var num_cycles = (cycles_fail * 2) + 1;
+
+    payload.autoboot = false;
+    payload.archive_on_delete = true;
```
misterdjules: Why is `archive_on_delete` meaningful?

src/vm/tests/test-docker.js
```diff
@@ -1121,3 +1123,194 @@ test('test docker VM with good zlog_max_size', function (t) {
...
+                t.ok((deltas[cycles_fail - 1] > deltas[cycles_fail]),
+                    deltas[cycles_fail - 1] + ' > ' + deltas[cycles_fail]);
+
+                // should be [0, 1, 2, ... 14]
+                for (i = 0; i < num_cycles; i++) {
+                    expected_counts.push(i);
+                }
+
+                // delays should be: [null, 200, 400 ... 12800, 200, ... 12800]
+                for (i = 0; i < cycles_fail; i++) {
```
misterdjules: Same as above: adding some of the content from the PR's description that describes the logic of the test would help understand why these values are expected.

src/vm/tests/test-docker.js
```diff
@@ -1121,3 +1123,194 @@ test('test docker VM with good zlog_max_size', function (t) {
...
+    var state = {brand: payload.brand};
+    var num_cycles = (cycles_fail * 2) + 1;
+
+    payload.autoboot = false;
+    payload.archive_on_delete = true;
+    payload.brand = 'lx';
+    payload.docker = true;
+    payload.image_uuid = vmtest.CURRENT_DOCKER_ALPINE_UUID;
+    // This cmd will 'exit 1' cycles_fail times, then sleep cycle_reset_delay
+    // seconds and exit 0, then repeat that pattern.
```
misterdjules: From the code below, I understand that it:

> 'exit 1' cycles_fail times, then sleep cycle_reset_delay seconds and exit 0

but I don't see where it "then repeat(s) that pattern".

src/vm/sbin/vmadmd.js
joshwilsdon: Yep, good catch.

src/vm/sbin/vmadmd.js
joshwilsdon: Yep, thanks.

src/vm/sbin/vmadmd.js
joshwilsdon: How about:

```
/*
 * The DOCKER_RUNTIME_DELAY_RESET parameter is used when restarting a Docker VM
 * according to its restart policy. As with Docker, we have an increasing delay
 * between each restart attempt for a container. The exception to this
 * increasing delay is when a container stays running for some amount of time.
 * The DOCKER_RUNTIME_DELAY_RESET is what controls this. If a docker VM was
 * running for longer than this many milliseconds when it stopped, we'll reset
 * the delay time between restarts. On Docker this is always 10 seconds, so we
 * use that here too.
 */
```

src/vm/sbin/vmadmd.js
joshwilsdon: Good point.

src/vm/sbin/vmadmd.js
joshwilsdon: Sure.

src/vm/sbin/vmadmd.js
joshwilsdon: We should probably at least log. At this point though, updateZoneStatus() doesn't call a callback, so it's not really clear what we should do when a VM.load fails for a single update. I'll file a ticket to look into whether we can do some sort of retry here.

src/vm/tests/test-docker.js
joshwilsdon: Good catch, we don't actually need this.

src/vm/tests/test-docker.js
joshwilsdon: it repeats the pattern because it deletes the files in /var/tmp with: `rm -f /var/tmp/*` at which point the `$(ls -1 /var/tmp | wc -l)` will go back to 0.

src/vm/tests/test-docker.js
joshwilsdon: Yep, I'll document this better.

src/vm/sbin/vmadmd.js
misterdjules: That looks good, thank you!

src/vm/tests/test-docker.js
misterdjules: Ah right, thanks!

misterdjules:
LGTM once the comments you agreed with are addressed.

PR #553, opened 2016-02-04
Remove unneeded 'an' for 'tags' is plural

jasperla:
Small textual tweak as "an additional network tags" isn't quite right and the rest of the sentence talks about "NICs".


trentm:
Thanks! Merged in commit 9c1caba7f2f9e5a08aadb06d8bf68779af80f302

PR #563, opened 2016-02-18
OS-5160  failure to create zone control file cascades into dumping core due to unchecked undefined callback

joshwilsdon:
This changes the directory creation so that it happens where we can do a retry.

It also adds a number of additional pieces of debugging information that will be available in cores.


misterdjules:
LGTM!

PR #572, opened 2016-03-02
use zone_state instead of state so we don't ignore provisioning VMs w…

joshwilsdon:
…hich might be running

### DIFF
--- a/src/vm/lib/metadata/agent.js
+++ b/src/vm/lib/metadata/agent.js
@@ -120,7 +120,8 @@ var sdc_fields = [
     'vnc_port',
     'zfs_io_priority',
     'zonepath',
-    'zonename'
+    'zonename',
+    'zone_state'
 ];
 var MAX_RETRY = 300; // in seconds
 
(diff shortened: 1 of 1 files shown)


src/vm/lib/metadata/agent.js
```diff
@@ -298,9 +299,9 @@ MetadataAgent.prototype.createServersOnExistingZones = function () {
             self.zones[zone.zonename] = zone;
             self.addDebug(zone.zonename, 'last_zone_load');
 
-            if (zone.state !== 'running') {
+            if (zone.zone_state !== 'running') {
                 self.log.debug('skipping zone ' + zone.zonename + ' which has '
-                    + 'non-running state: ' + zone.state);
+                    + 'non-running state: ' + zone.zone_state);
```
misterdjules: Should that be `'non-running zone_state:` instead to avoid confusion?

joshwilsdon: Sure, that seems reasonable.

misterdjules:
Besides my comment above about the debug message, LGTM.

PR #573, opened 2016-03-04
OS-5208 add support for stop -t

joshwilsdon:


rmustacc:
Is there a reason's there's no manual page update, even though the synopsis has been updated?

joshwilsdon:
@rmustacc I've now updated the man page

src/vm/man/vmadm.1m.md
```diff
@@ -337,6 +337,12 @@ tab-complete UUIDs rather than having to type them out for every command.
         If vmadmd sees a VM that has a transition but reaches the expiry before
         actually turning off, it re-sends the stop command with the -F option.
 
+        For Docker VMs, vmadm will send a SIGTERM to init and then wait some
+        number of seconds for the init process to exit. If it has not exited by
+        the timeout expiry, a SIGKILL will be sent. The default timeout is 10
+        seconds but this can be adjusted with the -t <timeout seconds> option.
+        For non-Docker VMs the -t option will do nothing.
```
rmustacc: I realize that history of vmadm is silently ignoring unknown arguments. But it seems like if I'm a user of other types of VMs the fact that it's silently ignored seems odd. Is there a reason we shouldn't make it an error to specify it if it's not actually supported for a given VM type?

joshwilsdon:
@rmustacc 

> I realize that history of vmadm is silently ignoring unknown arguments. But it seems like if I'm a user of other types of VMs the fact that it's silently ignored seems odd. Is there a reason we shouldn't make it an error to specify it if it's not actually supported for a given VM type?

In adding a failure here I remembered that timeout is actually used in KVM as well and settable with this same new -t option, so I've just pushed another commit 0c84d3c0dadf47fd578cf15ad7a56319660b31af which makes that more clear in the docs and also makes it an error to use -t for a non-KVM, non-docker zone.

If you try you'll get:

{code}
[root@headnode (coal) ~]# vmadm stop -t 20 b3dfa5fb-1d12-41b3-a0ca-0cae3b4cb229
Timeout option is only supported for KVM and docker containers.
[root@headnode (coal) ~]# echo $?
1
[root@headnode (coal) ~]#
{code}

PR #581, opened 2016-03-12
OS-5237 reuse the vmadmd watcher instead of spinning up a new zoneevent for e…

joshwilsdon:
…very VM.start()

By reusing the sysevent watcher, we avoid creating a separate zoneevent for each VM.start(). We also avoid the orphaning of these zoneevent processes which happened because the way we created them was not reentrant.


joshwilsdon:
Ran through the vmadm test suite:

```
#  TEST COMPLETE IN 4260 SECONDS, SUMMARY:
#
# PASS: 3291 / 3291
#
```

PR #594, opened 2016-04-01
OS-4897 detect and workaround missed or delayed sysevent messages

joshwilsdon:
This change adds the ability for metadata agent to detect when it has probably
missed sysevent messages or when they're significantly delayed.

It does this by polling the VM data for pid and boot_timestamp and comparing
these to the last object we loaded. When we see that a VM has changed its pid
and/or boot_timestamp but we've not seen an event, we assume we've missed a
sysevent that told us the VM has changed states. This will also happen when
delivery of a sysevent takes a long time.

In my testing I found that sometimes when the system is really busy it can
sometimes take at least 192 *seconds* for us to get a message that a VM has
started/stopped. I left this running with 128 VMs which were rebooted
repeatedly, as soon as they had completely come online. This ran for about 15
hours and did many thousands of reboots. None of the zones got stuck very long
waiting for a metadata socket.

I also tested by adding more zones (a total of 238) which caused the sysevents
to be significantly delayed and often dropped. Even in this setup the zones
continued to reboot without mdata:fetch going into maintenance.

Further testing has shown that the stop and start monitoring the agent has been doing
was based on a mistaken assumption that we couldn't reuse the socket across zone boots.
This assumption also infected OS-2253. So with this change we've removed the need to
remove the metadata sock from the zone on boot (brand statechange script) and instead
create a socket once and only recreate if the server gets closed. This greatly simplifies the
code here and reduces the amount of work metadata needs to do to keep itself up-to-date.

This change also improves the logging. Decreasing verbosity in some places and adding
additional details and timing information in others.

### DIFF
--- a/src/vm/lib/metadata/agent.js
+++ b/src/vm/lib/metadata/agent.js
@@ -34,7 +34,7 @@
  *   - attempts to create a metadata socket for all existing VMs on the CN
  *   - starts an interval timer so that we check every 5 minutes for VMs that
  *     have been deleted
- *   - starts a Zwatch watcher that calls a callback whenever a VM is started or
+ *   - starts a ZWatch watcher that calls a callback whenever a VM is started or
  *     stopped.
  *
  * When a zone starts, it attempts to create an appropriate server on a socket
@@ -72,6 +72,8 @@ var common = require('./common');
 var crc32 = require('./crc32');
 var execFile = require('child_process').execFile;
 var fs = require('fs');
+var getZoneinfo
+    = require('/usr/vm/node_modules/vmload/vmload-zoneinfo').getZoneinfo;
 var guessHandleType = process.binding('tty_wrap').guessHandleType;
 var net = require('net');
 var path = require('path');
@@ -85,6 +87,7 @@ var ZWatch = require('./zwatch');
 var sdc_fields = [
     'alias',
     'billing_id',
+    'boot_timestamp',
     'brand',
     'cpu_cap',
     'cpu_shares',
@@ -109,6 +112,7 @@ var sdc_fields = [
     'owner_uuid',
     'package_name',
     'package_version',
+    'pid',
     'quota',
     'ram',
     'resolvers',
@@ -124,6 +128,7 @@ var sdc_fields = [
     'zone_state'
 ];
 var MAX_RETRY = 300; // in seconds
+var ZONEADM_CHECK_FREQUENCY = (1 * 60 * 1000); // ms, check for deleted/running
 
 
 function zoneExists(zonename, callback) {
@@ -220,14 +225,37 @@ MetadataAgent.prototype.createZoneLog = function (type, zonename) {
     return (self.zlog[zonename]);
 };
 
-MetadataAgent.prototype.updateZone = function (zonename, callback) {
+MetadataAgent.prototype.updateZone =
+function updateZone(zonename, opts, callback) {
     var self = this;
     var log = self.log;
+
+    assert.string(zonename, 'zonename');
+    assert.object(opts, 'opts');
+    assert.func(callback, 'callback');
+
+    if (self.zones[zonename]) {
+        self.zones[zonename].updating = true;
+    }
 
     function shouldLoad(cb) {
         if (!self.zones.hasOwnProperty(zonename)) {
             // don't have a cache, load this guy
-            log.info('no cache for: ' + zonename + ', loading');
+            log.info({zonename: zonename},
+                'no cache for: ' + zonename + ', loading');
+            cb(true);
+            return;
+        }
+
+        if (opts.force) {
+            log.trace({zonename: zonename}, 'opts.force set, forcing reload');
+            cb(true);
+            return;
+        }
+
+        if (self.zones[zonename].zone_state === 'unknown') {
+            log.info({zonename: zonename},
+                'VM has unknown zone_state, forcing reload');
             cb(true);
             return;
         }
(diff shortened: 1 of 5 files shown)


src/vm/lib/metadata/agent.js
```diff
@@ -242,8 +303,9 @@ MetadataAgent.prototype.updateZone = function (zonename, callback) {
                 cb(false);
                 return;
             } else if (err) {
-                // fail open when we can't stat
-                log.error({err: err}, 'cannot fs.stat() ' + zonename);
+                // fail open (meaning: force reload) when we can't stat
+                log.error({err: err, zonename: zonename},
+                    'cannot fs.stat(), reloading');
                 cb(true);
```
misterdjules: I just realize that this callback pattern seems different than the `callback(errObject, output...)` that we use elsewhere. Why do we need to return a boolean value as the first parameter of the callback here and not an error object?

src/vm/lib/metadata/agent.js
```diff
@@ -254,26 +316,35 @@ MetadataAgent.prototype.updateZone = function (zonename, callback) {
...
             }
 
-            log.debug('using cache for: ' + zonename);
+            log.trace('using cache for: ' + zonename);
             cb(false);
         });
     }
 
     shouldLoad(function (load) {
+        var start_lookup_timestamp = (new Date()).getTime();
```
misterdjules: Should we use a monotonic clock (maybe `process.hrtime`) here instead of wall-clock time to avoid issues with the clock drifting?

misterdjules: It seems we use that value only for debugging, I thought we might have used it for something more critical. I still think it would be an improvement, but I'm not sure that's critical.

src/vm/lib/metadata/agent.js
```diff
@@ -315,18 +387,29 @@ MetadataAgent.prototype.createServersOnExistingZones = function () {
...
+            // created a socket for the zone. In case that happened, we ignore
+            // zones we've already got a connection for.
+            if (self.zoneConnections[zone.zonename]) {
+                cb();
+                return;
+            }
+
             if (zone.brand === 'kvm') {
                 self.startKVMSocketServer(zone.zonename, function (err) {
+                    created++;
```
misterdjules: Should we check for an error before incrementing `created`?

src/vm/lib/metadata/agent.js
```diff
@@ -315,18 +387,29 @@ MetadataAgent.prototype.createServersOnExistingZones = function () {
...
                     cb();
                 });
             } else {
                 self.startZoneSocketServer(zone.zonename, function (err) {
+                    created++;
```
misterdjules: Same as above.

src/vm/lib/metadata/agent.js
```diff
@@ -382,75 +499,123 @@ MetadataAgent.prototype.startDeletedZonesPurger = function () {
...
+    self.log.info('Setup timer to purge deleted zones every %d ms',
+        ZONEADM_CHECK_FREQUENCY);
+
+    setTimeout(_checkNewZones, MISSED_SYSEVENT_CHECK_FREQUENCY);
+    self.log.info('Setup timer to detect (missed) new zones every %d ms',
+        MISSED_SYSEVENT_CHECK_FREQUENCY);
+};
+
+MetadataAgent.prototype.handleZoneBoot = function handleZoneBoot(zonename) {
+    var self = this;
```
misterdjules: `assert.string(zonename, 'zonename')`?

src/vm/lib/metadata/agent.js
```diff
@@ -574,299 +763,243 @@ function (zonename, callback) {
...
+            zlog.warn({zonename: zopts.zone, err: error},
+                'failed to create sockpath directory');
+
+            // We were unable to create the directory but we have not yet
+            // created a real self.zoneConnections entry so we can just delete
+            // the placeholder and call callback. If the VM still exists and
+            // is running, we'll try again when we next poll w/
+            // _checkNewZones().
+            delete self.zoneConnections[zopts.zone];
+            callback();
```
misterdjules: It seems we're swallowing the error `error` here, but that at least some `createZoneSocket` call sites use the error passed in its callback, so there might be some inconsistency.

src/vm/lib/metadata/agent.js
```diff
@@ -1282,11 +1441,32 @@ MetadataAgent.prototype.makeMetadataHandler = function (zone, socket) {
 
         function addMetadata(cb) {
             var filename;
+            var start_add_metadata = (new Date()).getTime();
+
+            function _callCbAndLogTimer(err) {
```
misterdjules: Can we avoid duplicating `_callCbAndLogTimer` here and above?

src/vm/lib/metadata/agent.js
```diff
@@ -1282,11 +1441,32 @@ MetadataAgent.prototype.makeMetadataHandler = function (zone, socket) {
...
+                var elapsed = (new Date()).getTime() - start_add_metadata;
```
misterdjules: Maybe here using a monotonic clock instead of wall-clock time could prevent some inconsistent results (false positives and false negatives)?

In general I think we could use monotonic clock time everywhere in this PR where we currently use wall-clock time, but I'll let you make the call with regards to when and where to fix that.

src/vm/lib/metadata/agent.js
joshwilsdon: This shouldLoad() function's ultimately trying to answer a boolean question: "Should we load the VM now or not?", so the only argument the cb takes is a boolean. We could change all the calls to cb(null, [true|false]) if you think that's better. I just can't think of a case where we'd want to use the error. If stat failed here we want to reload.

src/vm/lib/metadata/agent.js
joshwilsdon: Sure. I can do that. Thanks for pointing this out.

src/vm/lib/metadata/agent.js
joshwilsdon: The only place this gets called is startZoneSocketServer() none of whose callbacks were doing anything with the error. However, with the changes you suggested wrt updating the "created" counter only when successful, this will change. So I'll pass the error back for that to work.

j4s0nmchr1st0s: Is it working?

misterdjules:
@joshwilsdon LGTM :)

src/vm/lib/metadata/agent.js
```diff
@@ -749,14 +840,15 @@ MetadataAgent.prototype.createKVMServer = function (zopts, callback) {
...
         delete self.zoneConnections[zopts.zone];
     });
 
     kvmstream.connect(zopts.sockpath);
 
-    zlog.info('listening on fd ' + kvmstream._handle.fd);
-    self.zoneConnections[zopts.zone].fd = kvmstream._handle.fd;
+    fd = kvmstream._handle.fd;
+    zlog.info('listening on fd %d', fd);
+    self.zoneConnections[zopts.zone].fd = fd;
```
misterdjules: Why do we explicitly create a `fd` property on the object referenced by `self.zoneConnections[zopts.zone]`? I would think that `self.zoneConnections[zopts.zone]` points to an object with a `conn` property that is an instance of `net.Socket`, which thus has a `._handle.fd` property. But I may be missing something.

joshwilsdon: As discussed directly with @misterdjules we do this to make it easier to pull out the handles for debugging. Also in the case that there's an error the handle is sometimes cleared immediately but before we've cleaned up entirely. If we crash in this state we might be able to use the fd we know the handler had to help diagnose what went wrong.

j4s0nmchr1st0s:
Joshua u in there?

PR #609, opened 2016-04-25
OS-5361 zfd device creation is async. Retry if they're not there initially.

joshwilsdon:
This changes so that whenever we open a zfd device from dockerinit we retry when the device does not exist yet. It also adds some debugging as to how long we waited in zfd_ready() and changes so that if the user has requested to open stdin and we can't open stdin after ZFD_OPEN_RETRIES, we fail instead of starting the user's init with /dev/null as stdin.


joshwilsdon:
This was implemented with 1bdb56d

PR #627, opened 2016-05-18
Support remote installation/recovery

jsonn:


rmustacc:
I'd really like to avoid adding yet more additional magical boot params that are required to be on the kernel comand line. Do we really have to add additional boot params here? We have a fixed length on it and the more we add, the more we risk hitting issues. The whole overload here is why boot time modules were created.

jsonn:
I have two slightly different use cases for this. In the office, I have a pre-configured USB disk and can put my static configuration, including root password, into the menu.lst for a "emergency recovery mode" option. For remote installation, the situation is more difficult. I can ask ISPs to fetch an off-the-shelve USB or ISO image and put it into the virtual media library. Even on a Skylake system, KVM works well enough to set the root_shadow value and the new run_sshd option. It wouldn't allow adding a new file. Having to create a file for this seems to be counterproductive.

That said, it would be useful to provide authorized_keys instead of the password, but for that using the command line is pretty much impossible.

PR #636, opened 2016-06-03
OS-5442 metadata agent cannot deal with stale sockets showing up post…

joshwilsdon:
… OS-4897

This changes the metadata agent to keep the fs.stat() info from a new socket when we create it. This way when we're checking for a stale socket we can also check the ino and dev fields and the ctime in order to detect metadata.sock files that have been swapped underneath us (from a zfs rollback for example).

This also adds a test case which reproduces the problem when not accompanied by the fix here.

Note: There is a bunch of other reformatting in that test case that goes along with this. That's because I've added the test file to the list of files we check with `make check` and I had to reformat the existing code in order to make it pass.

### DIFF
--- a/src/vm/lib/metadata/agent.js
+++ b/src/vm/lib/metadata/agent.js
@@ -234,29 +234,29 @@ function zoneExists(zonename, callback) {
  *        when the conn entry sockpath has a different fs.stat() signature
  *
  */
-function checkStaleSocket(conn, opts, callback) {
-    assert.optionalObject(conn, 'conn');
+function checkStaleSocket(zoneConn, opts, callback) {
+    assert.optionalObject(zoneConn, 'zoneConn');
     assert.object(opts, 'opts');
     assert.object(opts.log, 'opts.log');
     assert.func(callback, 'callback');
 
-    if (!conn || !conn.sockpath) {
-        // if there's no conn, it can't be stale
+    if (!zoneConn || !zoneConn.sockpath) {
+        // if there's no zoneConn, it can't be stale
         callback(null, false);
         return;
     }
 
-    assert.string(conn.sockpath, 'conn.sockpath');
-    assert.object(conn.sockstat, 'conn.sockstat');
+    assert.string(zoneConn.sockpath, 'zoneConn.sockpath');
+    assert.object(zoneConn.sockstat, 'zoneConn.sockstat');
 
-    fs.stat(conn.sockpath, function _onSockpathStat(err, stats) {
+    fs.stat(zoneConn.sockpath, function _onSockpathStat(err, stats) {
         var field;
         var fields = ['dev', 'ino']; // fields to compare in fs.Stats
 
         if (err) {
             if (err.code === 'ENOENT') {
                 opts.log.trace({
-                    sockpath: conn.sockpath
+                    sockpath: zoneConn.sockpath
                 }, 'ENOENT on sockpath: stale');
                 callback(null, true); // stale
                 return;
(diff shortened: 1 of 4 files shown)


src/vm/tests/test-snapshots.js
```diff
@@ -917,3 +980,153 @@ test('delete stopped joyent-minimal VM', function(t) {
...
+
+    t.ok(vmobj, 'have vmobj');
+    if (!vmobj) {
+        t.end();
+        return;
+    }
+
+    zonecontrol = vmobj.zonepath + '/root/.zonecontrol/metadata.sock';
+    try {
+        fs.mkdirSync(path.dirname(zonecontrol));
```
misterdjules: What happens if the dir exists, but not the file `metadata.sock` in it, is the test still working as expected, i.e does the "garbage" socket file still gets created?

src/vm/lib/metadata/agent.js
```diff
@@ -1121,7 +1177,20 @@ function createZoneSocket(zopts, callback) {
             zlog.info('listening on fd %d', fd);
             self.addDebug(zopts.zone, 'last_zsock_listen_success');
 
-            callback();
+            fs.stat(self.zoneConnections[zopts.zone].sockpath,
```
misterdjules: Is there a way to not duplicate that code with `createKVMServer`'s implementation?

src/vm/tests/test-snapshots.js
joshwilsdon: The socket actually gets created at the server.listen() below. And we will fail the t.ok() assertion there if fs.existsSync() doesn't find the socket at that point. So I think the test will work in this case.

src/vm/lib/metadata/agent.js
joshwilsdon: good point! I'll try to do this.

misterdjules:
LGTM.

src/vm/lib/metadata/agent.js
```diff
@@ -269,65 +269,76 @@ function checkStaleSocket(conn, opts, callback) {
...
         } else {
-            conn.sockstat = st;
+            zoneConn.sockstat = st;
         }
         callback(e);
     });
 }
 
+function closeZoneConnection(zoneConn) {
+    // .serverSocket is a net.Server for non-KVM
```
misterdjules: I think we should assert that `zoneConn` is an object. I would also suggest asserting that one but only one of `close` or `destroy` methods got called.

src/vm/lib/metadata/agent.js
```diff
@@ -228,36 +230,116 @@ function zoneExists(zonename, callback) {
...
+            // this as stale.
+            zoneConn.sockstat = {};
+        } else {
+            zoneConn.sockstat = st;
+        }
+        callback(e);
+    });
+}
+
+function closeZoneConnection(zoneConn) {
```
misterdjules: I think we should assert that `zoneConn` is an object. I would also suggest asserting that one but only one of `close` or `destroy` methods got called.

src/vm/tests/test-snapshots.js
misterdjules: What if the file already exists? It seems that ` fs.mkdirSync(path.dirname(zonecontrol));` would throw an error, but `vmobj.zonepath + '/root/.zonecontrol/metadata.sock'` would still exist, and thus `server.listen(zonecontrol, ...)` would throw an error too (`EADDRINUSE`).

joshwilsdon: Good point, I'll put the unlinkSync in its own try block and change so that we're checking the exception for the ones we expect and having the test fail if there's an exception that's not EEXIST for mkdir or ENOENT for unlink. Would that resolve this?

src/vm/lib/metadata/agent.js
joshwilsdon: Sure. Good point.

src/vm/tests/test-snapshots.js
misterdjules: That sounds good to me!

Issue #566, opened 2016-02-20
cyberpower usb ups not assigned correct driver


razamatan:
similar to https://github.com/joyent/smartos-live/issues/285

my ups is cyberpower brand.  here's the relevant prtconf -v output:

```
                input, instance #0
                    Hardware properties:
                        name='driver-minor' type=int items=1
                            value=00000000
                        name='driver-major' type=int items=1
                            value=00000002
                        name='usb-product-name' type=string items=1
                            value='CRCA102#BJ1'
                        name='usb-vendor-name' type=string items=1
                            value='CP1350PFCLCD'
                        name='usb-serialno' type=string items=1
                            value='CPS'
                        name='usb-raw-cfg-descriptors' type=byte items=34
                            value=09.02.22.00.01.01.00.c0.01.09.04.00.00.01.03.00.00.00.09.21.10.01.21
.01.22.ae.01.07.05.81.03.40.00.0a
                        name='usb-dev-descriptor' type=byte items=18
                            value=12.01.00.02.00.00.00.40.64.07.01.05.01.00.01.02.03.01
                        name='usb-release' type=int items=1
                            value=00000200
                        name='usb-num-configs' type=int items=1
                            value=00000001
                        name='usb-revision-id' type=int items=1
                            value=00000001
                        name='usb-product-id' type=int items=1
                            value=00000501
                        name='usb-vendor-id' type=int items=1
                            value=00000764
                        name='compatible' type=string items=9
                            value='usb764,501.1' + 'usb764,501' + 'usbif764,class3.0.0' + 'usbif764,cl
ass3.0' + 'usbif764,class3' + 'usbif,class3.0.0' + 'usbif,class3.0' + 'usbif,class3' + 'usb,device'
                        name='reg' type=int items=1
                            value=00000004
                        name='assigned-address' type=int items=1
                            value=00000003
```

so the classifier to use is probably **`usbif764,class3`**

and in /etc/driver_aliases, please add:

`ugen "usbif764,class3"`

leaberry:
I have ran into the same issue. It would be great if someone could just add the one line in /etc/driver_aliases so all us cyberpower ups users could have working configs. 

Also, until it's added what is the best way to add the line?

razamatan:
for now, i work around this by having a small startup script (that i call from a general smf service that runs all scripts in my startup scripts directory) that manually loads the driver by running this command:
`/usr/sbin/update_drv -a -i 'usbif764,class3' ugen`

PR #674, opened 2016-10-29
Add Cyberpower USB vendor class

leaberry:
This will make cyberpower usb UPS units attach to the ugen driver so apcupsd can use them. Fixes Issue #566


rmustacc:
Hmm. I'm wondering if we should actually just be identifying this class and leveraging some of the hid logic to have it always bind to ugen regardless.

leaberry:
I discovered after a number of hours that my cyberpower UPS won't work with apcupsd (wrong protocol) so I'm not as interested in getting this merged. But the guy from Issue #566 is probably interested.

I don't have any opinions on the best implementation.

ismell:
leaberry@ did you ever figure out a way of using your cyperpower UPS?

rmustacc:
@ismell Assuming that this is showing up as a hid device, all hid devices should show up with ugen nodes you can use now instead, so in theory the logic here should no longer be necessary.

leaberry:
@ismell I never got it working. The ugen part worked fine but apcupsd didn't support the cyberpower protocol for my UPS

razamatan:
i'm the guy from #566... if i undo my manual workaround, should things just work in the current release?

PR #733, opened 2017-07-13
fix build followed the wiki

pulux:
https://github.com/illumos/illumos-gate/commit/d1151f9bab097777251f26772ebcc50468637a12 breaks building

```
/usr/include/libcmdutils.h:41:2: error: #error "libcmdutils.h can only be used in a largefile compilation environment"
```

### DIFF
--- a/tools/tzcheck/Makefile
+++ b/tools/tzcheck/Makefile
@@ -29,6 +29,7 @@ SRC =		$(OBJECTS:%=.o%.c)
 CFLAGS =					\
 		-Wall -Wextra -Werror 		\
 		-std=c99 			\
+		-D_FILE_OFFSET_BITS=64          \
 		-D__EXTENSIONS__ 		\
 		-D_REENTRANT			\
 		-I../common			\


jclulow:
I believe @jasonbking is looking into this build failure (and a few other inappropriate uses of `libcmdutils` in `smartos-live.git`).  Jason?

jasonbking:
Yeah -- I was hoping to get to it once I got to a good stopping point with the IKEv2 stuff I've been neck deep in.  I think I got there this evening, so I'll take a look tomorrow and see if as was suggested, if we can have all the non illumos-joyent stuff in smartos-live avoid touching libcmdutils.  If that turns out to be problematic, we can fall back to just using _LARGEFILE64_SOURCE.

jasonbking:
It would also be good to understand exactly how you are building it as there are a few common targets (and a few not-so-common targets) -- just to be sure I'm not missing a path somewhere.

jasonbking:
I think we figured out why it was getting missed (and why it would build in some cases and not in others).  I have a fix that should address all of that in the process of building now.  Hopefully should have it tested today or tomorrow.

pulux:
Just followed https://wiki.smartos.org/display/DOC/Building+SmartOS+on+SmartOS and on `gmake live`, I hit the issue. The Zone has 4GB Ram and 60GB disk-space.

jasonbking:
I've just pushed the fix that has the users of custr use a private copy instead of the libcmdutils version.   Pulling the latest should now allow things to build on recent PIs

pulux:
works now as expected, ty.

PR #767, opened 2018-02-17
Enable unattended configuration for smartos when booting via USBImage

skippie81:
Hi,

Would like to enable the use of an answers.json file for configuration when booting form USB Disk.

So it is possible to use standalone smartos machines in a remote location.

Example where I use it:
Smartos machine located in a remote datacenter offering private hardware. I cannot use pxe boot as the datacenter offers a pxe solution for there rescue images.
So booting a rescue image they offer (Debian,Ubuntu, ... FreeBSD, ... ) I can get the latest USB image and dd it to a usb drive attached to the server.
But then I have no way of having an unattended configuration on boot from the USB drive. Now I need to request a KVM connection (that are sparse) and have to configure it manually via a remote kvm session ....

This fix will enable you to have answers.json file on the usb disk located in /private/answers.json
One key zpool_layout would be needed in the answers.json compared to the already avalable keys

examble:
```
{
  "skip_instructions": true,
  "config_console": "text",
  "skip_final_summary": true,
  "skip_final_confirm": true,
  "simple_headers": true,
  "hostname": "test-machine",
  "dns_search": "example.local",
  "dns_resolver1": "1.2.3.4",
  "dns_resolver2": "<default>",
  "ntp_host": "ntp.example.com",
  "admin_ip": "10.0.0.2",
  "admin_netmask": "255.255.255.0",
  "headnode_default_gateway": "10.0.0.1",
  "admin_nic": "08:00:27:08:85:4c",
  "zpool_layout": "<default>",
  "root_password": "someDefaultPassword"
}
```

### DIFF
--- a/overlay/generic/smartdc/lib/smartos_prompt_config.sh
+++ b/overlay/generic/smartdc/lib/smartos_prompt_config.sh
@@ -770,7 +770,13 @@ EOF
 		prmpt_str+=" To see a different configuration, type: 'raidz2', 'mirror', or 'default'.\n"
 		prmpt_str+=" To specify a manual configuration, type: 'manual'.\n\n"
 		print $prmpt_str
-		promptval "Selected zpool layout" "yes"
+
+    if [[ ${layout} == $(getanswer "zpool_layout") ]]; then
+      DISK_LAYOUT=/var/tmp/disklayout.json
+      return
+    fi
+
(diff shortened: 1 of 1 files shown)


skippie81:
Sorry, that i noticed the repo uses cr.joyent.us (Gerrit) after creating the pull request.
I'll close this one and follow instructions.

PR #782, opened 2018-05-23
Bump Maximum MTU to 9216

noahmehl:
Modern 10Gbe and above adapters and network gear support larger MTUs than 9000. This is important if users want to offer overlay networks of 9000 mtu, encapsulated in other protocols like vxlan, ipsec, vlan, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -450,6 +450,7 @@ clean:
 	(cd tools/mancheck && gmake clean)
 	(cd tools/mancf && gmake clean)
 	(cd tools/tzcheck && gmake clean)
+	(cd tools/ucodecheck && gmake clean)
 	(cd man/sdc && gmake clean)
 	rm -f 0-*-stamp 1-*-stamp
 


jclulow:
Ahoy!  I've had a look at this, and consulted with some of my colleagues as well.

I think it's important to understand where the original limit of 9000 came from.  In particular, we're trying to avoid allowing people to shoot themselves in the foot: if you have NICs today that allow, say, MTUs of 9216, that's great -- but in the future, you may end up on a part where either the hardware or the driver is contrained to an MTU of less than that (e.g., maybe only 9000).  If that happens you'll need to re-MTU your network to allow the new NICs to work.

That being said, it _is_ your computer, and we probably shouldn't be enforcing this right down at the bottom platform layer.  With that in mind, 9216 seems somewhat arbitrary -- we should probably just check that the value is an integer (as we do today) and that it's in the range from 1500 up to 65536 (the maximum value of a 16 bit unsigned integer).  Subsequent calls to `dladm` will fail if the value is too large for the driver/hardware anyway, and upstack Triton components can then be responsible for MTU caps as a site-wide policy decision.

As noted in the [Contributing section of the README](https://github.com/joyent/smartos-live#contributing, we don't generally use Github pull requests for most of our software (this repository included), but rather our Gerrit system at https://cr.joyent.us.  If you could create a change on our Gerrit, I can work on the review and approval process.  There's a [guide for new gerrit users](https://github.com/joyent/joyent-gerrit/blob/master/docs/user/README.md).

Thanks!

noahmehl:
@jclulow I hope I did this correctly: https://cr.joyent.us/#/c/4145/

trentm:
Hi. Some updates on this PR. It was moved to Gerrit (https://cr.joyent.us/4145) by request a while back. Since then, Joyent has moved away from cr.joyent.us and back to using GitHub PRs. While on Gerrit:

1. Cody opened TRITON-428 (https://jira.joyent.us/browse/TRITON-428 https://smartos.org/bugview/TRITON-428) for this a while back.

2. There was this discussion on gerrit (reproduced here because cr.joyent.us is going away):

## CR discussion

##### @melloc commented at 2018-06-05T22:02:06

> Patch Set 1:
>
> (2 comments)
>
> Thanks for doing this! Sorry for the confusion, and for not taking care of it sooner.
>
> Can you provide a comment here about how you've tested it? You were trying to set up a 9256 environment, right? Do these changes work in that environment?

##### @melloc commented at 2018-06-11T06:31:02

> Topic set to TRITON-428

##### Patch Set 1 code comments

> ###### /COMMIT_MSG#7 @melloc
>
> > Can we update the commit message to be:
> >
> >     TRITON-428 Increase maximum MTU to 65536
>
> ###### overlay/generic/lib/svc/method/net-physical#109 @melloc
>
> > can we make this line instead:
> >
> >     if ! [[ $mtu =~ ^[1-9][0-9][0-9][0-9][0-9]?$ ]]; then

noahmehl:
@trentm Thank you for keeping this alive!  Is there still any interest in completing this change?

trentm:
@noahmehl I'm not sure. I'll mention this PR to the OS team here. It will be up to them to respond here if they are able.

jlevon:
Hi Noah, did you have responses to Cody's review comments as well as his comments on testing? As far as I can tell that was the last change in this review

noahmehl:
@jlevon I'm fine with a max of 65536.  I haven't had any time to try adjusting the code base.

noahmehl:
Also, I don't even know where in the code base this is happening :( since the overlay folder no longer exists...

noahmehl:
well, I tried rebasing, and I totally failed :(

jlevon:
It's now been merged into illumos-joyent

noahmehl:
I think I've found it:

https://github.com/joyent/smartos-live/blob/0154453ee44db92440f7dfd682d8ffe2f9b63723/src/lib/sdc/network.sh#L151

noahmehl:
I have opened a new PR: https://github.com/joyent/smartos-live/pull/860

PR #793, opened 2018-07-19
sysinfo CPU core terminology is confusing

esproul:
The term "core" is confusing here because it is used to describe
both the physical CPU count ("CPU Physical Cores") and the total
number of vCPUs ("CPU Total Cores"). IMO, "physical cores" should
refer to the actual core count, but that is not something that is
currently enumerated.  That may be a followup change.

I propose renaming the physical CPU count as "CPU Physical
Processors" to better communicate what this number represents.

### DIFF
--- a/man/src/sysinfo.1m.md
+++ b/man/src/sysinfo.1m.md
@@ -75,7 +75,7 @@ making changes.
 
     GZ and NGZ
 
-  "CPU Physical Cores"
+  "CPU Physical Processors"
 
     Number of physical CPUs in this host.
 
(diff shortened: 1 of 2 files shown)


jclulow:
Unfortunately the output of `sysinfo` is a cross-consolidation interface, so we can't easily change the names of existing keys.  We can definitely add new keys, and we can make the documentation better describe what's going on, though.

jlevon:
Hi @esproul, thanks for your changes. As it happens, I'm resolving this for other reasons, and adding new keys that will hopefully be much clearer. As @jclulow mentions, we can't easily remove the existing ones, but at least there will be better named ones to refer to.

PR #858, opened 2019-11-01
OS-8030 vminfod does not offer a way to tell if a vm is fully provisioned

twhiteman:
Test suite is still passing after this change:
```
#
#  TEST COMPLETE IN 4956 SECONDS, SUMMARY:
#
# PASS: 5145 / 5145
#
# log files available in: /tmp/vmtest.1572625585.35075
```


sjorge:
Does this also work when a vm goes through a reprovision?

twhiteman:
Moved to Gerrit (https://cr.joyent.us/#/c/7073).

PR #877, opened 2019-11-21
OS-8051 OS-8032 blocks ./configure if git output is too big

danmcd:
Tested by running ./configure on my standard 80x24 terminal window.


jlevon:
small nit: need "is too big" in the bug title

PR #881, opened 2019-12-10
TOOLS-2392 node-contract should not be an ignored jr-manifest.json entry

timfoster:


timfoster:
Trent mentioned on chat that the ignore list is there to say that the ignored repositories aren't relevant for this manifest, not ignored across all combined jr-manifests that we know about. As such, this PR should be closed.

PR #885, opened 2019-12-16
OS-7698 Add mkzpool -e for creating encrypted zpools

jasonbking:


src/node_modules/zfs.js
```diff
@@ -124,22 +125,35 @@ zpool.status = function (pool, callback) {
  * zpool create command, including log devices, cache devices, and hot spares.
  * The input is an object of the form produced by the disklayout library.
  */
-zpool.create = function (pool, config, force, callback) {
+zpool.create = function (pool, config, force, encrypt, callback) {
+    var cmd;
```
mgerdts: Do we now have a weird mixture of spaces and tabs?

src/node_modules/zfs.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Joyent, Inc.  All rights reserved.
+ * Copyright 2019 Joyent, Inc.  All rights reserved.
```
mgerdts: `Copyright 2019 Joyent, Inc.`

man/man/man1m/mkzpool.1m
```diff
@@ -0,0 +1,41 @@
+.Dd November 13, 2019
```
bahamat: Date needs to be updated

jasonbking: The date on man pages is usually the date of last modification (that's how it is for all of the illumos man pages).

bahamat: Discussed with @jasonbking, this is accurate as to when the change was authored, so I call it good.

PR #886, opened 2019-12-16
OS-8028 Have sysinfo report pool recovery configurations

jasonbking:


src/sysinfo
```diff
@@ -691,6 +691,18 @@ function get_setup()
     fi
 }
 
+function get_recovery_configs()
+{
+    for line in $(kbmadm recovery list -p 2>/dev/null); do
```
mgerdts: Initialize `has_recovery` and add a comment explaining its significance outside of this function.  Ugh to the lack fo scoped variables throughout this file.

src/sysinfo
```diff
@@ -691,6 +691,18 @@ function get_setup()
...
+        fields=(${line//:/ })
+        cfg=${fields[0]}
+        cfguuid=${fields[1]}
+        [[ -n "$cfguuid" ]] && \
```
mgerdts: An `if` statement would be clearer.

src/sysinfo
```diff
@@ -910,6 +922,21 @@ if [[ ${ZONENAME} == "global" ]]; then
         echo '  "Zpool Size in GiB":' ${Zpool_size}','
     fi
 
+    if [[ -n "has_recovery" ]]; then
```
mgerdts: `ouput_parseable()` should have something like this too.

jasonbking: That'd probably better to be fixed with OS-7859 since the rest of the zpool stuff is missing from the parseable output (I actually have a change w/ most of that done already, but is missing the recovery bits since it predates them).  Either way the `sysinfo -p` isn't needed by anything for EDAR (the agents use the JSON output).

src/sysinfo
```diff
@@ -693,13 +693,19 @@ function get_setup()
...
+    # encryption should never emit this property.
+    has_recovery=
+
     for line in $(kbmadm recovery list -p 2>/dev/null); do
         fields=(${line//:/ })
         cfg=${fields[0]}
         cfguuid=${fields[1]}
-        [[ -n "$cfguuid" ]] && \
+        if [[ -n "$cfguuid" ]]; then
             printf -v "Zpool_Recovery_${cfg}" "%s" "$cfguuid" && \
```
mgerdts: No need for "&& \"

src/sysinfo
```diff
@@ -693,13 +693,19 @@ function get_setup()
...
             has_recovery=1
+        fi
      done
```
mgerdts: 5 spaces before done?

PR #894, opened 2020-01-09
OS-7868 need kbmadm and kbmd

jasonbking:


tools/jr-manifest.json
```diff
@@ -53,6 +53,13 @@
                 "mg": "platform"
             }
         },
+        {
+            "name": "kbmd",
+            "labels": {
+                "release": true,
+                "mg": "platform"
+            }
+        },
```
trentm: this file change LGTM

PR #899, opened 2020-01-30
TOOLS-2428 deliver modern CTF tools

jlevon:


Makefile
```diff
@@ -360,6 +360,10 @@ strap-cache-location:
 strap-cache:
 	$(ROOT)/tools/build_strap -c -j $(MAX_JOBS) -a $(ADJUNCT_TARBALL)
 
+# build a CTF tools tarball
```
jlevon: I think this file should probably go in the bits / artifacts directory somehow?

timfoster: The way to do this, I think would be to introduce a "strap-release" target, with a "strap-publish" and "ctf-publish" target, and then reuse the common "platform-publish" target to actually post the binaries up to Manta. This would mean we could drop some of the custom mput logic in the strap-cache build Jenkins shell script and have that in the Makefile instead.

See eg. "triton-smartos-release:" for an example of what we're doing elsewhere.

We probably ought to override "PLATFORM_DEBUG_SUFFIX" in the Jenkins job so that the strap and tools content doesn't go to the standard /Joyent_Dev/public/builds/platform directory.

(Best not to archive them with the Jenkins artifacts as those get thrown away periodically)

jlevon:
I think this is ready for review now

PR #903, opened 2020-02-11
OS-8110 fix strap cache download

jlevon:


jlevon:
> This seems fine, but I'm confused why we don't have manta-client tools available? I mean, bits-upload.sh needs them present elsewhere in the build, so if they're not available here, we're only going to fall over later trying to find them, aren't we?

We shouldn't require them as the strap cache is for developers - and most of those won't be doing bits-upload.

PR #908, opened 2020-02-21
TOOLS-2433 SmartOS platform image manifests should use "smartos" as OS instead of "other"

kusor:


kusor:
Platform build with the changes succeeded. This is the image: 2f17772b-4164-497a-8c47-1f98705bd1c5 available in experimental channel. So far, so good.

PR #911, opened 2020-03-05
OS-8046 add Jenkinsfiles to smartos repos

timfoster:


Jenkinsfile
```diff
@@ -0,0 +1,273 @@
...
+                '<dd>clobber Illumos before each build [default: no]</dd>\n' +
+                '<dt>-d</dt>\n' +
+                '<dd>build Illumos in DEBUG mode only [default: no]</dd>\n' +
+                '<dt>-h</dt>\n' +
+                '<dd>this message</dd>\n' +
+                '<dt>-p gcc4</dt>\n' +
+                '<dd>primary compiler version [default: gcc7]</dd>\n' +
+                '<dt>-P password</dt>\n' +
+                '<dd>platform root password [default: randomly chosen]</dd>\n' +
+                '<dt>-r</dt>\n' +
```
jlevon: this can be removed from this list since it's always over-ridden anyway

Jenkinsfile
```diff
@@ -0,0 +1,273 @@
...
+                '  <li>SmartOS iso image</li>\n' +
+                '  <li>SmartOS usb image</li>\n' +
+                '  <li>SmartOS vmware image</li>\n' +
+                '  <li>SmartOS Changelog file</li>\n' +
+                '</ul></p>'
+        )
+        booleanParam(
+            name: 'BUILD_STRAP',
+            defaultValue: false,
+            description: 'This parameter declares whether to build a new ' +
```
jlevon: This could probably do with a bit more clarification. We *always* build the strap, so this is perhaps not a good name for the variable. BUILD_STRAP_CACHE would be better,
and it should clarify that this is about building - and uploading - the strap cache used by devs, not the proto.strap that's used by Jenkins builds.

Jenkinsfile
timfoster: Ok, I can remove '-r'. Does anyone use the -P option, would you prefer if that stayed?

Jenkinsfile
timfoster: That makes sense. I'll change that, updated the wording and modify illumos-extra accordingly.

PR #913, opened 2020-03-12
OS-8131 changelog generation needs work after Jenkinsfile conversion

timfoster:


tools/build_changelog
```diff
@@ -59,7 +59,11 @@ function build_section
 	directory=$1
 	lastbranch=$2
 	cd $directory || fail "Unable to cd to $directory"
-	description=${3:-$(basename $PWD)}
+	if [[ "$directory" = "." ]]; then
```
jlevon: This is an obscure enough corner that it merits a comment IMHO

timfoster: Yep, will do.

PR #914, opened 2020-03-18
OS-8139 strap cache build not triggered due to incorrect variable name

timfoster:


timfoster:
I've added testing notes to OS-8139, looks like it's working correctly now.

PR #915, opened 2020-03-19
OS-8140 platform build workspaces overwhelming jenkins agents

timfoster:


timfoster:
I've added additional testing notes to the jira ticket - I think this is ready for IA at this point.

PR #917, opened 2020-03-23
TRITON-2090 would like a way to exclude disks from CN zones pool

jlevon:


src/disklayout.js
```diff
@@ -20,21 +20,25 @@ function
 usage()
 {
 	console.log('usage: ' + process.argv[0] +
-	    ' [-c] [-f file] [-s spares] [-w width] [<layout>]');
+	    ' [-c] [-f file] [-s spares] [-w width] [-i disk,...] [<layout>]');
```
YanChii: I see a typo here (`s/-i/-e/`).

jlevon:
Hey @timfoster mind looking at IA here?

PR #921, opened 2020-03-31
TOOLS-2454 jenkins-joylib should allow mm notification comments

timfoster:


timfoster:
I'm in two minds as to whether we should drop the 'os' notification for 'pipeline complete' That'll be relatively noisy for the os channel, as it'll include every PR update notification. Any strong opinions either way?

jlevon:
I don't think it's a big deal personally, there's not that much building going on

timfoster:
Added bahamat, to seek IA on this change once it has built.

PR #924, opened 2020-04-09
OS-8158 Support NVMe as device model for bhyve

sjorge:
Counter part for joyent/illumos-joyent#287 as this needs changes in both locations, progress and motivation tracked in the other PR


sjorge:
I dropped the commit mentioning this only works on windows, @papertigers also confirmed it works on linux with the firmware we ship.

PR #931, opened 2020-05-07
OS-8164 smartos-live projects infrastructure should add refspecs for pull requests

timfoster:


configure
```diff
@@ -109,9 +109,24 @@ function checkout_project
...
+			git -C ${dir} fetch origin ${branch}:${branch_desc}
+
+			# If the above fails, see if the project appears to be a GitHub
+			# PR-xxx branch name, and if so, add the relevant refspec for
+			# pull requests.
+			if [[ $? -ne 0 ]]; then
+				PR_BRANCH=$(echo "${branch}" | sed -E -e 's/^PR-[0-9]+$//g')
+				if [[ -z "${PR_BRANCH}" ]]; then
+					echo "Branch appears to be from a PR, adding GitHub refspec"
+					PR_NUMBER=$(echo "${branch}" | awk -F- '{print $2}')
```
jlevon, line 121: You can do this:

PR_NUMBER=$(echo "${branch}" | sed -nEe 's/^PR-([0-9]+)$/\1/p')
if [[ -n "${PR_NUMBER}" ]]; then
    ...
fi

to avoid a sed/grep then awk.

jlevon: For the echo, can we output the refspec we're going to use for debugging purposes?

timfoster: That's much nicer, and yes, I'll add that echo. Thanks for the review!

PR #932, opened 2020-05-11
OS-8170 size SmartOS USB for smaller cards

ahayes:
Some memory cards such as the Kingston 2GB SDCard that ships in Dell servers have fewer than 2GB of space for writing the SmartOS boot image. I have lowered the image size by 60MB to accommodate cards that fall short.

I have made minimal changes so as to avoid touching the "proforma" workflow or options and have tested building and running from this image on aforementioned cards.

Discussion here: https://smartos.topicbox.com/groups/smartos-discuss/Tecbe189dfeb6ebfc/the-partition-size-is-fixed-1-86g-in-usb-stick-can-it-be-enlarged

### DIFF
--- a/tools/build_boot_image
+++ b/tools/build_boot_image
@@ -45,7 +45,7 @@ bi_make_iso=0
 bi_proforma_only=0
 bi_nocleanup=0
 bi_imgsz_gb="2"
-bi_imgsz=$(( $bi_imgsz_gb * 1000000000 ))
+bi_imgsz="1940000256"
 bi_wsroot=
 bi_lofi_blkdev=
 bi_lofi_rawdev=
@@ -373,7 +373,7 @@ if [[  $bi_make_iso == 1 ]]; then
 	exit 0
 fi
 
-print "Creating $bi_imgsz_gb GB image at $bi_tmpdir/smartos.usb ... \c"
+print "Creating $bi_imgsz byte image at $bi_tmpdir/smartos.usb ... \c"
 create_image $bi_tmpdir $bi_imgsz $bi_tmpdir/smartos.usb
 print "done"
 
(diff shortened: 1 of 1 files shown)


tools/build_boot_image
```diff
@@ -45,7 +45,7 @@ bi_make_iso=0
 bi_proforma_only=0
 bi_nocleanup=0
 bi_imgsz_gb="2"
-bi_imgsz=$(( $bi_imgsz_gb * 1000000000 ))
+bi_imgsz="1940000256"
```
jlevon, line 48: Please add a comment as to why we have such a weird value.

tools/build_boot_image
```diff
@@ -45,7 +45,7 @@ bi_make_iso=0
 bi_proforma_only=0
 bi_nocleanup=0
 bi_imgsz_gb="2"
```
jlevon, line 47: Now this no longer reflects reality, please can you:

 - replace "bi_imgsz_gb" with something like "bi_proforma_prefix=""
 - fix up the -p handling to set that to ${OPTARG}gb and bi_imgsz to $OPTARG * 1000000000

tools/build_boot_image
ahayes: I committed a comment just above the initialization of variables section to explain the value. It seemed to be in keeping with how other documentation was organized. Let me know if you would prefer it on the line directly above or following on the same line instead.

tools/build_boot_image
ahayes: Thanks. New commit has replaced variable with new suggested name and fixed up the spots it was used.

jlevon:
@timfoster could you look at IA please?

timfoster:
Yep, looks fine. Thanks for fixing this.

jlevon:
Thanks!

PR #933, opened 2020-05-14
OS-8174 lay groundwork for GCC9

jlevon:


src/Makefile
```diff
@@ -177,19 +177,35 @@ C_CHECK_TARGETS=\
...
 
+#
+# The libuuid Node add-on follows the same pattern as EXPAT.NODE.
+#
+UUID_OBJDIR =	node-libuuid/build/Release
+UUID_SRCDIR =	node-libuuid/src
+UUID.NODE =	$(UUID_OBJDIR)/uuid.node
+
+$(UUID.NODE) :	NODE_WARN_FLAGS += -Wno-unused -Wno-write-strings
+$(UUID.NODE) :  NODE_WARN_FLAGS += -Wno-sign-compare
```
jasonbking, line 194: Nit: seems misaligned with the surrounding definitions (I only mention it since you fixed the EXPAT one above).

jlevon: done

PR #941, opened 2020-06-04
TRITON-2137 expunge v8plus users (fix gcc4)

jlevon:


jlevon:
Yes, I force pushed that branch for the fix.

I did a build to check the fix, can you IA?

timfoster:
Great, thanks for the confirmation.

PR #942, opened 2020-06-12
OS-8198 Allow bootable-off-ZFS-pool ("standalone") SmartOS

danmcd:
Allow a bootable `zones` pool, OR if manual pool configuration generates a `standalone` pool, use that.  Bootable pools can be on either BIOS or UEFI.

This is a WIP.  As indicated in the installer comments, the initial thought is to have PIs live in $BOOTPOOL`/boot/platform-YYYYxxxxx/` directories, and have $BOOTPOOL`/boot/platform` symlink to the desired PI.

Needs lots more work, including better install-time mechanisms, and a tool not unlike `sdcadm platform` to deal with PIs.  (One thought I had was `piadm`(1M), which future `sdcadm platform` could use on the Head Node.)

### DIFF
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -24,8 +24,8 @@ piadm(1M) -- Manage SmartOS Platform Images
     referred to here as a PI-stamp.  One can see it in uname(1M):
 
         smartos-build(~)[0]% uname -a
-	SunOS smartos-build 5.11 joyent_20200602T173751Z i86pc i386 i86pc
-	smartos-build(~)[0]% 
+        SunOS smartos-build 5.11 joyent_20200602T173751Z i86pc i386 i86pc
+        smartos-build(~)[0]% 
 
     The PI-stamp for this system's Platform Image is `20200602T173751Z`.
 
@@ -35,17 +35,29 @@ piadm(1M) -- Manage SmartOS Platform Images
 
         - The SmartOS `unix` kernel
 
-	- The SmartOS boot archive containing kernel modules, libraries,
-	  commands, and more.
+        - The SmartOS boot archive containing kernel modules, libraries,
+          commands, and more.
 
         - A manifest and hash.
 
-	- A file containing the PI-stamp.
+        - A file containing the PI-stamp.
 
     The SmartOS loader(5) will find a path to a Platform Image on the
     bootable ZFS pool, and will load `unix` and then the boot archive.
 
-    Platform images are supplies by a gzipped tarball containing the above.
+    Platform images are supplied by either a gzipped tarball containing the
+    above. Or inside an ISO image file which contains the above AND the boot
+    image as well (see below).
+
+## BOOT IMAGES
+
+    In addition to platform images, the loader(5) also has a directory
+    structure containing the loader itself and its support files.  These are
+    stamped as well with PI stamps, but are distinct from the contents of a
+    gzipped PI tarball.  Often, a PI can use an older Boot Image to boot
+    itself without issue.  Occasionally, however, a PI will have Boot Image
+    changes also that need to accompany it.
+
 
 ## COMMANDS
 
@@ -56,38 +68,65 @@ piadm(1M) -- Manage SmartOS Platform Images
 
         Activate a Platform Image for the next boot, on a specified ZFS pool
         if there are more than one bootable pools imported.  It is up to the
-        administrator to know which pool the system will actually boot.
+        administrator to know which pool the system will actually boot.  If a
+        boot image with the specified PI-stamp is unavailable, a warning will
+        be issued but the new PI will be activated anyway.
 
         `activate` and `assign` are synonyms, for those used to other
         distros' `beadm`, or Triton's `sdcadm platform`, respectively.
 
-      bootable [-d|-e] [ZFS-pool-name]
+      bootable [-d|-e [-i <source>]] [ZFS-pool-name]
 
         Query or upgrade a ZFS pool's bootable status.  With no arguments,
         the status of all imported pools will be queried.  -d will disable a
-        pool from being bootable, and -e will enable one.  As mentioned
-        earlier, it is up to the administrator to know which pool the system
-        will actually boot.
+        pool from being bootable, and -e will enable one.  If the -i flag
+        specifies an installation source, see below in the `install`
+        subcommand, it will be used.  Lack of -i is equivalent to `-i media`.
+        As mentioned earlier, it is up to the administrator to know which
+        pool the system will actually boot. Unlike install, this command will
+        always attempt to install a corresponding boot image as well.
 
-	Some pools can only be bootable from an older BIOS system, while
-	other can also be bootable from UEFI systems.  The `bootable`
-	subcommand will indicate this.
+        Some pools can only be bootable from an older BIOS system, while
+        other can also be bootable from UEFI systems.  The `bootable`
+        subcommand will indicate this.
 
-      install <PI-stamp, PI-tarball, PI-tarball-URL, "latest"> [ZFS-pool-name]
+      install <source> [ZFS-pool-name]
 
-        Installs a new Platform Image into the bootable pool.  If there are
-        more than one bootable pools, a pool name will be required.
-        piadm(1M) requires a Platform Image gzipped tar file.  If a PI-stamp
-        or the word "latest" is supplied, the well-known public SmartOS PI
-        repository will be queried with the specified PI-stamp.
+        Installs a new Platform Image into the bootable pool.  If the source
+        also contains the boot image (like an ISO does), the Boot Image will
+        also be installed, if available.  If there are more than one bootable
+        pools, a pool name will be required.  piadm(1M) requires a Platform
+        Image source.  That source can be:
+
+          - A PI-stamp, which will consult the well-known SmartOS PI
+            repository for an ISO image.  This requires network reachability
+            and working name resolution.
+
+          - The word "latest", which will consult the well-known SmartOS PI
+            repository for the latest ISO image.  This requires network
+            reachability and working name resolution.
+
+          - The word "media", which will attempt to find a mountable optical
+            media (CD or DVD) or USB-key with SmartOS on it.  The SmartOS
+            installer uses this keyword.
+
+          - An ISO image file path.
+
+          - A PI gzipped tarball file path.  NOTE this source does not have
+            a boot image in it.
+
+          - A URL to either one of an ISO image or a gzipped PI tarball.
 
       list [ZFS-pool-name]
 
-        Lists the available platform images on bootable pools.
+        Lists the available platform images (and boot images) on bootable
+        pools.
 
       remove <PI-stamp> [ZFS-pool-name]
 
-        The opposite of `install`, and only accepts a PI-stamp.
+        The opposite of `install`, and only accepts a PI-stamp.  If a boot
+        image exists with the specified PI-stamp, it will also be removed
+        unless it is the only boot image available.
 
 
 
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -241,8 +241,8 @@ The following exit values are returned:
          An error occurred, but no change was made
 
      2
-	 A fatal error occurred, and there may be partial change or
-	 other residual files or directories.
+         A fatal error occurred, and there may be partial change or
+         other residual files or directories.
 
      3
          A corrupt environment on what is supposed to be a bootable pool.
@@ -273,3 +273,7 @@ The following exit values are returned:
     UEFI.  A hand-partitioned GPT disk may be able to be bootable with both
     BIOS and UEFI, and can have some of its other GPT parititions used for
     other purposes.
+
+    If a bootable pool's boot image or platform image becomes corrupt, even
+    if it's `zones`, a machine can still be booted with a USB stick, CD-ROM,
+    or other method of booting SmartOS, and recovery methods can occur.
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -73,13 +73,13 @@ piadm(1M) -- Manage SmartOS Platform Images
 	other can also be bootable from UEFI systems.  The `bootable`
 	subcommand will indicate this.
 
-      install <PI-stamp, PI-tarball, PI-tarball-URL> [ZFS-pool-name]
+      install <PI-stamp, PI-tarball, PI-tarball-URL, "latest"> [ZFS-pool-name]
 
         Installs a new Platform Image into the bootable pool.  If there are
         more than one bootable pools, a pool name will be required.
         piadm(1M) requires a Platform Image gzipped tar file.  If a PI-stamp
-        is supplied, the well-known public SmartOS PI repository will be
-        queried with the specified PI-stamp.
+        or the word "latest" is supplied, the well-known public SmartOS PI
+        repository will be queried with the specified PI-stamp.
 
       list [ZFS-pool-name]
 
@@ -104,9 +104,6 @@ The following exit values are returned:
 
      1
          An error occurred.
-
-     2
-         Invalid Platform Image
 
 
 ## SEE ALSO
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -5,8 +5,7 @@ piadm(1M) -- Manage SmartOS Platform Images
 ## SYNOPSIS
     /usr/sbin/piadm [-v | -vv] <command> [command-specific arguments]
 
-    piadm activate <PI-stamp> [ZFS-pool-name]
-    piadm assign <PI-stamp> [ZFS-pool-name]
+    piadm activate|assign <PI-stamp> [ZFS-pool-name]
     piadm bootable
     piadm bootable [-dr] <ZFS-pool-name>
     piadm bootable -e [ -i <source> ] <ZFS-pool-name>
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -2,7 +2,7 @@ piadm(1M) -- Manage SmartOS Platform Images
 ===========================================
 
 ## SYNOPSIS
-    /usr/sbin/piadm [-v] <command> [command-specific arguments]
+    /usr/sbin/piadm [-v | -vv] <command> [command-specific arguments]
 
 ## DESCRIPTION
 
@@ -101,7 +101,11 @@ drwxr-xr-x   4 root     root           5 Jul 15 04:12 platform-20200714T195617Z
 
 ## COMMANDS
 
-    The following commands and options are supported:
+    The piadm(1M) command will produce more verbose output if run with -v, or
+    with -vv will produce both -v output and enable the shell's -x flag,
+    which produces output of all of the commands run in the piadm(1M) script.
+
+    piadm(1M) commands and options are:
 
       activate <PI-stamp> [ZFS-pool-name]
       assign <PI-stamp> [ZFS-pool-name]
@@ -234,7 +238,14 @@ The following exit values are returned:
          Successful completion.
 
      1
-         An error occurred.
+         An error occurred, but no change was made
+
+     2
+	 A fatal error occurred, and there may be partial change or
+	 other residual files or directories.
+
+     3
+         A corrupt environment on what is supposed to be a bootable pool.
 
 
 ## SEE ALSO
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -255,9 +255,21 @@ The following exit values are returned:
 ## NOTES
 
     Many ZFS pool types are not allowed to be bootable.  The system's BIOS or
-    EFI must locate a bootable disk on a bootable pool in order to boot.
+    UEFI must locate a bootable disk on a bootable pool in order to boot.
+    Future work in illumos will enable more ZFS pool types to be bootable,
+    but for now a ZFS pool should be a single-level-vdev pool, namely one of:
+
+      - Single disk
+      - Mirror
+      - RaidZ (any parity)
 
     SmartOS still loads a ramdisk root with a read-only /usr filesystem, even
-    with a bootable pool.
+    when booted from a bootable pool.  This means a bootable pool that isn't
+    the SmartOS `zones` pool receives relatively few writes unless it it used
+    for some other purpose as well.
 
-    <Notes of ZFS pools>
+    A bootable pool created without the -B option, but using whole disks,
+    will be BIOS bootable thanks to space for the MBR, but not bootable with
+    UEFI.  A hand-partitioned GPT disk may be able to be bootable with both
+    BIOS and UEFI, and can have some of its other GPT parititions used for
+    other purposes.
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -187,7 +187,7 @@ drwxr-xr-x   4 root     root           5 Jul 15 04:12 platform-20200714T195617Z
 [root@smartos ~]# piadm bootable
 standalone                     ==> non-bootable
 zones                          ==> non-bootable
-[root@smartos ~]# piadm bootable -e -i latest standalone
+[root@smartos ~]# piadm -v bootable -e -i latest standalone
 Installing PI 20200701T231659Z
 Platform Image 20200701T231659Z will be loaded on next boot,
     with a new boot image,
@@ -207,17 +207,15 @@ PI STAMP           BOOTABLE FILESYSTEM            BOOT BITS?   NOW   NEXT
 [root@smartos ~]# piadm list
 PI STAMP           BOOTABLE FILESYSTEM            BOOT BITS?   NOW   NEXT
 20200714T195617Z   standalone/boot                next         yes   yes
-[root@smartos ~]# piadm install https://example.com/PIs/platform-20200715T192200Z.tgz
+[root@smartos ~]# piadm -v install https://example.com/PIs/platform-20200715T192200Z.tgz
 Installing https://example.com/PIs/platform-20200715T192200Z.tgz
         (downloaded to /tmp/tmp.Bba0Ac)
 Installing PI 20200715T192200Z
-umount: warning: /tmp/tmp.XbaqBc/mnt not in mnttab
-umount: /tmp/tmp.XbaqBc/mnt not mounted
 [root@smartos ~]# piadm list
 PI STAMP           BOOTABLE FILESYSTEM            BOOT BITS?   NOW   NEXT
 20200714T195617Z   standalone/boot                next         yes   yes
 20200715T192200Z   standalone/boot                none         no    no
-[root@smartos ~]# piadm activate 20200715T192200Z
+[root@smartos ~]# piadm -v activate 20200715T192200Z
 Platform Image 20200715T192200Z will be loaded on next boot,
     WARNING:  20200715T192200Z has no matching boot image, using
     boot image  20200714T195617Z
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -1,0 +1,124 @@
+piadm(1M) -- Manage SmartOS Platform Images
+===========================================
+
+## SYNOPSIS
+    /usr/sbin/piadm [-v] <command> [command-specific arguments]
+
+## DESCRIPTION
+
+    Historically, SmartOS booted off of a USB key or a read-only media like
+    CD-ROM.  The copy and version of the SmartOS software on one of these
+    media is called a Platform Image.  A Platform Image is described in
+    detail in the next section. 
+
+    piadm(1M) manages multiple copies of Platform Images on a bootable ZFS
+    pool, allowing easier updates to Platform Images and maintaining multiple
+    Platform Images on a single boot media.  The method and implementation of
+    SmartOS booting does not change vs. a USB key or CD-ROM, but merely uses
+    a bootable ZFS pool as the source of the Platform Image, which can be the
+    SmartOS `zones` pool if it is a bootable pool.
+
+## PLATFORM IMAGES
+
+    A SmartOS Platform Image (PI) is identified by creation timestamp,
+    referred to here as a PI-stamp.  One can see it in uname(1M):
+
+        smartos-build(~)[0]% uname -a
+	SunOS smartos-build 5.11 joyent_20200602T173751Z i86pc i386 i86pc
+	smartos-build(~)[0]% 
+
+    The PI-stamp for this system's Platform Image is `20200602T173751Z`.
+
+    The Platform Image is a directory containing:
+
+        - A directory structure in a format used by loader(5).
+
+        - The SmartOS `unix` kernel
+
+	- The SmartOS boot archive containing kernel modules, libraries,
+	  commands, and more.
+
+        - A manifest and hash.
+
+	- A file containing the PI-stamp.
+
+    The SmartOS loader(5) will find a path to a Platform Image on the
+    bootable ZFS pool, and will load `unix` and then the boot archive.
+
+    Platform images are supplies by a gzipped tarball containing the above.
+
+## COMMANDS
+
+    The following commands and options are supported:
+
+      activate <PI-stamp> [ZFS-pool-name]
+      assign <PI-stamp> [ZFS-pool-name]
+
+        Activate a Platform Image for the next boot, on a specified ZFS pool
+        if there are more than one bootable pools imported.  It is up to the
+        administrator to know which pool the system will actually boot.
+
+        `activate` and `assign` are synonyms, for those used to other
+        distros' `beadm`, or Triton's `sdcadm platform`, respectively.
+
+      bootable [-d] [-e] [ZFS-pool-name]
+
+        Query or upgrade a ZFS pool's bootable status.  With no arguments,
+        the status of all imported pools will be queried.  -d will disable a
+        pool from being bootable, and -e will enable one.  As mentioned
+        earlier, it is up to the administrator to know which pool the system
+        will actually boot.
+
+	Some pools can only be bootable from an older BIOS system, while
+	other can also be bootable from UEFI systems.  The `bootable`
+	subcommand will indicate this.
+
+      install <PI-stamp, PI-tarball, PI-tarball-URL> [ZFS-pool-name]
+
+        Installs a new Platform Image into the bootable pool.  If there are
+        more than one bootable pools, a pool name will be required.
+        piadm(1M) requires a Platform Image gzipped tar file.  If a PI-stamp
+        is supplied, the well-known public SmartOS PI repository will be
+        queried with the specified PI-stamp.
+
+      list [ZFS-pool-name]
+
+        Lists the available platform images on bootable pools.
+
+      remove <PI-stamp> [ZFS-pool-name]
+
+        The opposite of `install`, and only accepts a PI-stamp.
+
+
+
+## EXAMPLES
+
+
+
+## EXIT STATUS
+
+The following exit values are returned:
+
+     0
+         Successful completion.
+
+     1
+         An error occurred.
+
+     2
+         Invalid Platform Image
+
+
+## SEE ALSO
+
+    zpool(1M), loader(5)
+
+## NOTES
+
+    Many ZFS pool types are not allowed to be bootable.  The system's BIOS or
+    EFI must locate a bootable disk on a bootable pool in order to boot.
+
+    SmartOS still loads a ramdisk root with a read-only /usr filesystem, even
+    with a bootable pool.
+
+    <Notes of ZFS pools>
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -25,7 +25,7 @@ piadm(1M) -- Manage SmartOS Platform Images
 
         smartos-build(~)[0]% uname -a
         SunOS smartos-build 5.11 joyent_20200602T173751Z i86pc i386 i86pc
-        smartos-build(~)[0]% 
+        smartos-build(~)[0]%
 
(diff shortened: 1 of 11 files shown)


src/lib/sdc/usb-key.sh
```diff
@@ -133,6 +135,45 @@ function unmount_usb_key()
...
+# Evil twins of the {,un}mount_usb_key functions.  These are specific
+# to ISO/hsfs (aka. {C,DV,B}D-ROM) disks.  We may be able to factor-out
+# even more common bits from the usb_key functions, but not today.
+
+function mount_ISO
+{
+	local mnt=`extract_mountpath $1`
+
+	# XXX KEBE SAYS JUST TRUST IT FOR NOW...
+	for disk in `disklist -r`; do
```
jasonbking, line 147: Consider using `diskinfo -p`

src/lib/sdc/usb-key.sh
```diff
@@ -133,6 +135,45 @@ function unmount_usb_key()
...
+		mount -F hsfs /dev/dsk/${disk}s0 $mnt
```
jasonbking, line 148: If this fails, will it print out an error?
`fstyp` could be useful here to weed out non-hsfs devices -- wouldn't need to use  `2>/dev/null` to quiet things we just want to skip, but still produce error messages for things we are interested in.

src/lib/sdc/usb-key.sh
danmcd: Doesn't seem to detect CD-ROM/DVD-ROM drives.
```
[root@smartos-bios ~]# mount -F hsfs /dev/dsk/c2t1d0s0 /mnt
[root@smartos-bios ~]# df -h /mnt/.
Filesystem             Size   Used  Available Capacity  Mounted on
/dev/dsk/c2t1d0s0    554.11M 554.11M          0   100%    /mnt
[root@smartos-bios ~]# diskinfo -p
TYPE    DISK                    VID      PID              SIZE          RMV SSD
SCSI    c1t0d0                  VMware,  VMware Virtual S 34359738368   no  yes
SCSI    c1t1d0                  VMware,  VMware Virtual S 34359738368   no  yes
SATA    c2t0d0                  VMware   Virtual SATA Hard Drive 34359738368   no  yes
SATA    c2t2d0                  VMware   Virtual SATA Hard Drive 34359738368   no  yes
[root@smartos-bios ~]# 
```

src/lib/sdc/usb-key.sh
danmcd: Yes.  The function returns 0 (success) mid-loop, and the bottom of the function is the failure case.

src/lib/sdc/usb-key.sh
jasonbking: That's unfortunate.. oh well..

src/lib/sdc/usb-key.sh
```diff
@@ -133,6 +135,45 @@ function unmount_usb_key()
 	umount "$mnt"
 }
 
+# Evil twins of the {,un}mount_usb_key functions.  These are specific
+# to ISO/hsfs (aka. {C,DV,B}D-ROM) disks.  We may be able to factor-out
+# even more common bits from the usb_key functions, but not today.
+
+function mount_ISO
```
sjorge, line 142: nit: lowercase ISO? from a quick glance all function names are lowercase

sjorge:
> Needs lots more work, including better install-time mechanisms, and a tool not unlike `sdcadm platform` to deal with PIs. (One thought I had was `piadm`(1M), which future `sdcadm platform` could use on the Head Node.)

This is for new installs only in the current state right? Will this future tool also be able to convert exsting cdrom/usb booting nodes to having it on the zones pool?

src/lib/sdc/usb-key.sh
danmcd: ISO in all caps is easier to grep.  Not sure it's worth it.

src/lib/sdc/usb-key.sh
danmcd: I do need to ax the KEBE comment here, however.

danmcd:
> This is for new installs only in the current state right? Will this future tool also be able to convert exsting cdrom/usb booting nodes to having it on the zones pool?

Right now, this is only about new installs.  Experiments have demonstrated that a pool not created with `zpool create -B` CAN be induced to boot from BIOS (either as standalone, or even to iPXE for a Triton CN), but they cannot be induced to boot from UEFI w/o pool reformatting/etc.  I have in my `piadm` notes a `piadm bootable` which could affect such a change, but that's all still very much up in the air.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1339,6 +1437,51 @@ sed -e "s|^root:[^\:]*:|root:${root_shadow}:|" /etc/shadow > /usbkey/shadow \
...
+
+	# Determine which disk(s) to install things in.
+	# Then for each disk:
+	# 	installboot -m -b....
+	#
+	# We created the pool, so we can assume
+	# `zpool create -B` generates things such that for disk X, Xs0 == ESP,
+	# Xs1 == data-for-BOOTPOOL
+	for a in \
+	`zpool list -v ${BOOTPOOL} | egrep 'c[0-9]+' | awk '{print $1}'`; do
```
bahamat, line 1474: This is probably safe in this context, but in general somewhat dangerous.

Prefer:

```
mapfile -t boot_devices < <(zpool list -v "${BOOTPOOL}" | grep -E 'c[0-9]+' | awk '{print $1}')
for a in "${boot_devices[@]}"; do
```

Also, should the regex be more specific? `c[0-9A-F]+t[0-9A-F]+d[0-9A-F]+'`? (I don't know if the `c` or `d` values can go up to A-F, but `t` can.)

src/mkzpool.js
```diff
@@ -18,7 +18,7 @@ fatal(msg)
 function
 usage()
 {
-	console.log('usage: ' + process.argv[0] + '[-f] <pool> <file.json>');
+	console.log('usage: ' + process.argv[0] + '[-Bef] <pool> <file.json>');
```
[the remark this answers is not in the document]
jasonbking: No -- it was just missed during the EDAR putback.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -969,8 +996,14 @@ create_zpool()
 
 	# If this is not a manual layout, then we've been given
 	# a JSON file describing the desired pool, so use that:
-	mkzpool -f $pool $layout || \
-	    fatal "failed to create pool ${pool}"
+	# Try and make a bootable one first.
+	mkzpool -B -f $pool $layout
```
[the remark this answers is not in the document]
jasonbking: I'd say yes -- this alone doesn't make the zpool bootable -- it just leaves room for the EFI bootable partition on the drives in the pool, so there's always the option of making the pool bootable in the future (even if the operator doesn't elect to do so initially).

I don't think the small amount of space (a few MB) it takes from each disk for the EFI partition should present any space concerns (if you're that tight on space where it makes a difference, there are likely other more critical problems that are going to occur).

src/mkzpool.js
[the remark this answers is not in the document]
bahamat: 👍

src/smartdc/lib/smartos_prompt_config.sh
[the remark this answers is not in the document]
bahamat: Ok, that makes sense. Thanks.

src/lib/sdc/usb-key.sh
```diff
@@ -133,6 +135,45 @@ function unmount_usb_key()
...
+{
+	local mnt=`extract_mountpath $1`
+
+	# XXX KEBE SAYS JUST TRUST IT FOR NOW...
+	for disk in `disklist -r`; do
+		mount -F hsfs /dev/dsk/${disk}s0 $mnt
+		if [[ $? != 0 ]]; then
```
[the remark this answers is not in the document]
danmcd: I changed it to be `-ne` to match other instances in this script.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -739,6 +741,27 @@ printdisklayout()
...
+promptyesno()
+{
+    val=""
+    while [ -z "$val" ]; do
+	printf "$1 [$2]: "
+	read val
+	[ -z "$val" ] && val="$2"
+
+	# Normalize to lowercase
+	val=`echo $val | sed 's/A-Z/a-z/g'`
```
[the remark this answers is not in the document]
danmcd: Will fix.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1234,6 +1306,29 @@ your own zpool.\n"
...
+SmartOS can boot off either the \"zones\" zpool, or a dedicated \"standalone\"
+zpool in lieu of a USB stick or a CD-ROM.  Enter 'yes' if you wish to try and
+make this SmartOS zpool self-booting.\n"
+
+	
+	if [[ $(getanswer "skip_instructions") != "true" ]]; then
+	    printf "$message"
+	fi
+
+	promptyesno "Boot SmartOS from disk" $ondisk
```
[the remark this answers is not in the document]
danmcd: Will fix.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1005,6 +1038,45 @@ create_zpools()
 	touch /${SYS_ZPOOL}/.system_pool
 }
 
+copy_installmedia()
+{
+	# Try the USB key first, quietly...
+	mount_usb_key /mnt > /dev/null 2>&1 
+	if [[ $? != 0 ]]; then
+	    # If that fails, try mounting the ISO.
+	    mount_ISO /mnt || fatal "Odd, can't find install media!"
```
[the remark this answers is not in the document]
danmcd: Will fix.  (Moral equivalent of a KEBE comment... :) )

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1339,6 +1437,51 @@ sed -e "s|^root:[^\:]*:|root:${root_shadow}:|" /etc/shadow > /usbkey/shadow \
...
+    # its default, which is NO bootsize.
+    zpool get bootsize ${BOOTPOOL} | grep -q -w default
+    if [[ $? == 0 ]]; then
+	fatal "\nDesired bootable pool $BOOTPOOL was not created with -B"
+    fi
+
+    # Create BOOTPOOL/boot and set bootfs.
+    zfs create ${BOOTPOOL}/boot || fatal "Cannot create boot filesystem"
+    zpool set bootfs=${BOOTPOOL}/boot ${BOOTPOOL}
+    if [[ $? != 0 ]]; then
```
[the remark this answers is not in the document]
danmcd: Going to use the `-ne` syntax in place already instead.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: I don't see `mapfile` readily available for the SmartOS installer.  (It's not in smartos-live/src either.)

As for the regex, old-school IDE (like on some VM/cloud environments) have cNdN.  I figure matching on c{num} was okay, but if the controller counts in hex, yeah, we need A-F too.

bahamat: It’s a bash shell builtin in v4, so will be available.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1339,6 +1438,51 @@ sed -e "s|^root:[^\:]*:|root:${root_shadow}:|" /etc/shadow > /usbkey/shadow \
...
+    # Reality check $BOOTPOOL was created with -B.
+    # Easiest way to do this is to check for the `bootsize` property not
+    # its default, which is NO bootsize.
+    zpool get bootsize ${BOOTPOOL} | grep -q -w default
+    if [[ $? -eq 0 ]]; then
+	fatal "\nDesired bootable pool $BOOTPOOL was not created with -B"
+    fi
+
+    # Create BOOTPOOL/boot and set bootfs.
+    zfs create ${BOOTPOOL}/boot || fatal "Cannot create boot filesystem"
```
jasonbking, line 1453: We might want to explicitly add `-o encryption=off` here. That way it doesn't need to care about the rest of the pool.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: Wow.  I thought bash would answer "shell built-in" to `which mapfile`.  :(  Thanks.  Will update later today.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: Good idea.  Will update later today.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: Check now.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: Check now.

src/lib/sdc/usb-key.sh
bahamat: Here, on line 144 prefer `local mnt=$(extract_mountpath "$1")` over backticks.

On line 146, same `mapfile` thing as below in `smartos_config_prompt.sh`

    mapfile -t disks < <(disklist -r)
    for a in "${disks[@]}"; do
        ...
    done

danmcd:
This most recent commit gets us to a testable place. There are install-time corner cases that may bite, pools created with slices are iffy, and whole-disk pools created without -B *can* boot on BIOS systems, but not on EFI (and `piadm bootable` can attempt to discern the difference).

danmcd:
Opening up for proper review.  Still subject to change in face of testing and test results, but happy enough with it to expose it more more scrutiny.

danmcd:
Now with proper bug id:  OS-8198

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,263 @@
...
+
+## NOTES
+
+    Many ZFS pool types are not allowed to be bootable.  The system's BIOS or
+    EFI must locate a bootable disk on a bootable pool in order to boot.
+
+    SmartOS still loads a ramdisk root with a read-only /usr filesystem, even
+    with a bootable pool.
+
+    <Notes of ZFS pools>
```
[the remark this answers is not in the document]
danmcd: Damn.  Will fix.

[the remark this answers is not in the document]
danmcd: Should be better now.

danmcd:
Time permitting, I think this is closer to ready now.

src/piadm.sh
```diff
@@ -54,50 +54,65 @@ vecho() {
     fi
 }
 
-declare -a allbootable
-declare -a numbootable
+declare bootfs
+declare -a allbootfs
+declare -a numbootfs
```
jasonbking, line 59: Is this really an array, or is just `allbootfs`?

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -273,3 +271,7 @@ The following exit values are returned:
     UEFI.  A hand-partitioned GPT disk may be able to be bootable with both
     BIOS and UEFI, and can have some of its other GPT parititions used for
     other purposes.
+
+    If a bootable pool's boot image or platform image becomes corrupt, even
+    if it's `zones`, a machine can still be booted with a USB stick, CD-ROM,
+    or other method of booting SmartOS, and recovery methods can occur.
```
jasonbking, line 277: 'recovery methods can occur' sounds a bit like recovery can spontaneously happen sometimes. Maybe reword it a bit?

src/piadm.sh
```diff
@@ -109,13 +124,10 @@ piname_present_get_bootfs() {
...
-    echo $bootfs
 }
 
-# Use "-k" for now until we ship CAs with the Platform Image again.
-CURL="curl -k"
+# Defined as a variable in case we need to add parameters (like -s) to it.
+CURL="curl -s"
 
 # Well-known source of SmartOS Platform Images
 URL_PREFIX=https://us-east.manta.joyent.com/Joyent_Dev/public/SmartOS/
```
jasonbking, line 133: Not necessary in v1, but maybe consider making this a conditional assignment (e.g. `DEFAULT_URL_PREFIX=https://us-east....; URL_PREFIX=${URL_PREFIX:-${DEFAULT_URL_PREFIX}}`) to allow people to be able to specify a custom URL in a future change.

src/piadm.sh
danmcd: It is an array of all of the bootable filesystems.  It also is the list of bootable pools.

man/usr/share/man/man1m/piadm.1m.md
danmcd: Will fix.

src/piadm.sh
jasonbking: Isn't that `allbootfs`, what about `numbootfs`?

src/piadm.sh
danmcd: Good idea, I think I'll do that now (but leave it undocumented).

src/piadm.sh
danmcd: Doh!  Will fix.

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+
+update_boot_sectors() {
+    pool=$1
+    bootfs=$2
+
+    # XXX WARNING -- illumos#12894 will allow slogs.  We will need to
+    # alter the generation of boot_devices accordingly.
+    # Generate the pool's boot devices now, in case we did something
+    # hyper-clever for the pool.  s1 may be created, but not yet PCFS...
+    mapfile -t boot_devices < <(zpool list -v "$pool" | \
```
jlevon, line 392: it's unfortunate that we kidn of have to use zpool list, but can it at least be zpool list -vHP zones | awk -F\\t '{print $2}'

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+        grep -E 'c[0-9]+' | awk '{print $1}' | sed -E 's/s[0-9]+//g')
+
+    # Reality check the pool was created with -B.
+    # First way to do this is to check for the `bootsize` property not
+    # its default, which is NO bootsize.
+    zpool get bootsize $pool | grep -q -w default
```
jlevon, line 398: same comment as elsewhere

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+    if [[ $? -eq 0 ]]; then
+	# No bootsize is a first-cut test.  It passes if the pool was
+	# created with `zpool create -B`. There's one other that needs
+	# to be peformed, because some bootable pools are manually
```
jlevon, line 402: nit, "performed"

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+	# configured to share slices with other functions (slog,
+	# l2arc, dedup):
+
+	# Use fstyp to confirm if this is a stealth EFI booter...
```
jlevon, line 406: what does stealth mean in this context?

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+	type=$(fstyp /dev/dsk/${boot_devices[0]}s0)
+	if [[ "$type" == "pcfs" ]]; then
+	    # If we detect PCFS on s0, it's LIKELY an EFI System Partition that
```
jlevon, line 409: do we even need or want to support a pool without "bootsize" ?

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+
+}
+
+activate() {
+    pistamp=$1
+    piname_present_get_bootfs $pistamp $2
+    pool=$(echo $bootfs | awk -F/ '{print $1}')
+
+    cd /$bootfs
+    bootstamp=$(file -h platform | awk '{print $5}' | sed 's/\.\/platform-//g')
```
jlevon, line 451: readlink means you're not parsing english text

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+#
+
+#
+# Copyright 2020 Joyent, Inc.
+#
+
+. /lib/sdc/usb-key.sh
+
+err() {
+    echo $1
```
jlevon, line 21: err() etc should be going to stderr

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+		echo "$pistamp is the current set of boot binaries.  Please"
+		echo "activate another pi using 'piadm activate <other-PI-stamp>' first."
+		usage
+	    fi
+	    /bin/rm -rf boot-$pistamp
+	fi
+
+	/bin/rm -rf platform-$pistamp
+    else
+	echo "$pistamp is not a stamp for a PI on pool $pool"
```
jlevon, line 517: generally, errors should be going to stderr

src/piadm.sh
```diff
@@ -0,0 +1,741 @@
...
+	    # Check for pcfs partition on pool disks.
+	    mapfile -t boot_devices < <(zpool list -v "${pool}" | \
+		grep -E 'c[0-9]+' | awk '{print $1}')
+	    for a in "${boot_devices[@]}"; do
+		noslice=$(echo $a | sed -E 's/s[0-9]+//g')
+		tdir=`mktemp -d`
+		# Assume that s0 on the physical disk would be where the EFI
+		# System Partition (ESP) lives.  A pcfs mount, ALONG WITH a
+		# check for a bootx64.efi executable, can confirm/deny it. Do
+		# this instead of just checkint for bootsize because we can
```
jlevon, line 680: "checking"

src/smartdc/lib/smartos_prompt_config.sh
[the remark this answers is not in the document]
danmcd: Now all in piadm.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1339,6 +1437,51 @@ sed -e "s|^root:[^\:]*:|root:${root_shadow}:|" /etc/shadow > /usbkey/shadow \
...
 cp -rp /etc/ssh /usbkey/ssh || fatal "failed to set up preserve host keys"
 
+if [ $ondisk == "yes" ]; then
+    printf "%-56s" "Creating self-bootable $BOOTPOOL pool... "
+
+    # Reality check $BOOTPOOL was created with -B.
+    # Easiest way to do this is to check for the `bootsize` property not
+    # its default, which is NO bootsize.
+    zpool get bootsize ${BOOTPOOL} | grep -q -w default
+    if [[ $? == 0 ]]; then
```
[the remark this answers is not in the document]
danmcd: Now all in piadm

src/smartdc/lib/smartos_prompt_config.sh
danmcd: Now all in piadm.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1339,6 +1437,51 @@ sed -e "s|^root:[^\:]*:|root:${root_shadow}:|" /etc/shadow > /usbkey/shadow \
 
 cp -rp /etc/ssh /usbkey/ssh || fatal "failed to set up preserve host keys"
 
+if [ $ondisk == "yes" ]; then
```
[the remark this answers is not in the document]
danmcd: Now in piadm.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
+piadm(1M) -- Manage SmartOS Platform Images
+===========================================
+
+## SYNOPSIS
+    /usr/sbin/piadm [-v | -vv] <command> [command-specific arguments]
+
```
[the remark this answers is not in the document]
danmcd: Will fix, using zfs or zpool as good examples.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+## DESCRIPTION
+
+    Historically, SmartOS booted off of a USB key or a read-only media like
+    CD-ROM.  The copy and version of the SmartOS software on one of these
+    media is called a Platform Image.  A Platform Image is described in
+    detail in the next section.
+
```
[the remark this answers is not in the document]
danmcd: Adding "The piadm(1M) utility enables and manages the ability to instead boot directly off of a ZFS pool." to the highlighted paragraph.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+## COMMANDS
+
+    The piadm(1M) command will produce more verbose output if run with -v, or
+    with -vv will produce both -v output and enable the shell's -x flag,
+    which produces output of all of the commands run in the piadm(1M) script.
+
+    piadm(1M) commands and options are:
+
+      activate <PI-stamp> [ZFS-pool-name]
```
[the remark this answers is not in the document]
danmcd: Added already in the summary, and will do so here.  (Need to specify that one can insert -v or -vv between piadm and the command, as well.)

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+        subcommand, it will be used.  Lack of -i is equivalent to `-i media`.
+        As mentioned earlier, it is up to the administrator to know which
+        pool the system will actually boot. Unlike install, this command will
+        always attempt to install a corresponding boot image as well.
+
+        The -r flag will refresh a bootable pool's MBR and/or ESP.  This is
+        especially useful on mirror or raidz pools that have new devices
+        attached.
+
+        Some pools can only be bootable from an older BIOS system, while
```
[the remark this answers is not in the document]
danmcd: Will fix.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+    Future work in illumos will enable more ZFS pool types to be bootable,
+    but for now a ZFS pool should be a single-level-vdev pool, namely one of:
+
+      - Single disk
+      - Mirror
+      - RaidZ (any parity)
+
+    SmartOS still loads a ramdisk root with a read-only /usr filesystem, even
+    when booted from a bootable pool.  This means a bootable pool that isn't
+    the SmartOS `zones` pool receives relatively few writes unless it it used
```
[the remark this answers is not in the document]
danmcd: Will fix

src/piadm.sh
danmcd: Yes, we do, because of BIOS-era pools.

```
[root@moe (kebecloud) ~]# piadm list
PI STAMP           BOOTABLE FILESYSTEM            BOOT BITS?   NOW   NEXT  
20200728T010227Z   zones/boot                     next         yes   yes  
[root@moe (kebecloud) ~]# piadm bootable
zones                          ==> BIOS 
[root@moe (kebecloud) ~]# zpool get bootsize
NAME   PROPERTY  VALUE     SOURCE
zones  bootsize  -         default
[root@moe (kebecloud) ~]# 
```

danmcd: My corner case here is the "double-duty ssd" case where it's sliced into pieces for boot and slog-or-l2arc.

src/piadm.sh
danmcd: Hmmm, I can use an entry in that directory, can't I?  `bootstamp=$(cat platform/etc/version/platform)`  Will fix.

src/piadm.sh
danmcd: Will fix.  Also will introduce `eecho` for cases like ones in the next comment.

src/piadm.sh
danmcd: `eecho` to the rescue.

src/piadm.sh
danmcd: Will fix.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1234,6 +1272,32 @@ your own zpool.\n"
...
+SmartOS can boot off either the \"zones\" zpool, or a dedicated named
+zpool in lieu of a USB stick or a CD-ROM.  Enter 'yes' if you wish to try and
+make a SmartOS zpool self-booting.\n"
+
+	
+	if [[ $(getanswer "skip_instructions") != "true" ]]; then
+	    printf "$message"
+	fi
+
+	promptyesno "Boot SmartOS from a zpool" $boot_from_zpool
```
jlevon, line 1295: This still seems to be forcing interactive usage?

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -1,15 +1,26 @@
 piadm(1M) -- Manage SmartOS Platform Images
 ===========================================
 
+
 ## SYNOPSIS
     /usr/sbin/piadm [-v | -vv] <command> [command-specific arguments]
 
+    piadm activate <PI-stamp> [ZFS-pool-name]
+    piadm assign <PI-stamp> [ZFS-pool-name]
```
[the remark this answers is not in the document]
danmcd: Will fix.

src/piadm.sh
```diff
@@ -17,39 +17,44 @@
 
 . /lib/sdc/usb-key.sh
 
+eecho() {
+    echo $@ > /dev/stderr
```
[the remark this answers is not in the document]
danmcd: like `echo $@ 1>&2` ?  Will fix.

src/piadm.sh
```diff
@@ -150,13 +155,13 @@ mount_installmedia() {
...
-		echo "Can't find install media: USB stick errors:"
-		echo ""
-		cat $tfile
-		echo ""
-		echo "ISO errors"
-		echo ""
-		cat $tfile2
+		eecho "Can't find install media: USB stick errors:"
+		eecho ""
+		cat $tfile > /dev/stderr
```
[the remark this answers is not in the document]
danmcd: Will also fix.

src/piadm.sh
```diff
@@ -150,13 +155,13 @@ mount_installmedia() {
...
+		eecho ""
+		eecho "ISO errors"
+		eecho ""
+		cat $tfile2 > /dev/stderr
```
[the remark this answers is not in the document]
danmcd: Will also also fix.

src/node_modules/zfs.js
```diff
@@ -125,20 +125,24 @@ zpool.status = function (pool, callback) {
...
  */
-zpool.create = function (pool, config, force, encrypt, callback) {
-	var cmd;
-	var args;
+zpool.create = function (pool, config, force, encrypt, efi, callback) {
+        var cmd;
+        var args;
 
-	if (arguments.length === 3) {
+    if (arguments.length === 3) {
```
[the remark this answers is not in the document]
danmcd: Other recent work in all-JS repos has me trained that all JS *MUST* be space-indented.  I might've used the big "M-x untabify" hammer to get it JS-style happy.

src/piadm.sh
```diff
@@ -0,0 +1,773 @@
...
+
+err() {
+    eecho $1
+    exit 1
+}
+
+fatal() {
+    eecho
+    if [[ -n "$1" ]]; then
+	eecho "ERROR: $1"
```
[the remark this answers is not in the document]
danmcd: NOW I get what's going on.  This is a weird emacs-indentation rule that works for cstyle (which has tab as 1st level, and assumes tab == 8 spaces, and has 4-space half-tabs as an indent rule) applied to shell scripts.  1st level is half-tab/4-spc, followed by 8 spaces, but HEY, that's a tab.  I will either use the big-untabify-hammer(no-tabs) OR go all-tabs(no-half-tabs/spaces).  Sorry for being slow on this.

src/smartdc/lib/smartos_prompt_config.sh
danmcd: $boot_from_zpool is set to "no" at the very top, so just hitting RETURN here will answer "no".  If "hitting RETURN a new time" is "forcing interactive usage" then yes, I'm adding one (two if you answer "yes") more question.

src/piadm.sh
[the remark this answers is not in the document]
jlevon: don't need the 1, just >&2

src/piadm.sh
danmcd: Sorry, I forgot -P is a guarantee of parsability.

danmcd:
I will also be picking one of tabs or spaces, now that I get this isn't cstyle.

src/smartdc/lib/smartos_prompt_config.sh
jlevon: Please check the rest of the script: it should be possible to supply a config file such that it automatically installs. That's what you're breaking by requiring any input. Look at getanswer() is you're still not clear

danmcd:
Tested new installation script that should address @jlevon's feedback:

```
7.) Install time pool testing (post-jlevon CR).
7a.) Hit RETURN (no bootable pool, can't even EFI-activate it) (PASSED)
7b.) Specify "zones" with auto-layout (PASSED)
7c.) Specify "standalone" with auto-layout, should complain (PASSED, maybe
     direct piadm output to /dev/null ?!?)
7d.) Create "zones" and "standalone" manually, should print choices and you
     can/should specify one of them (PASSED)
```

danmcd:
Tested new installation script that should address @jlevon's feedback:

```
7.) Install time pool testing (post-jlevon CR).
7a.) Hit RETURN (no bootable pool, can't even EFI-activate it) (PASSED)
7b.) Specify "zones" with auto-layout (PASSED)
7c.) Specify "standalone" with auto-layout, should complain (PASSED, maybe
     direct piadm output to /dev/null ?!?)
7d.) Create "zones" and "standalone" manually, should print choices and you
     can/should specify one of them (PASSED)
```

jlevon:
I think you need to push the new smartos prompt changes that are no longer interactive?

danmcd:
> I think you need to push the new smartos prompt changes that are no longer interactive?

DAMN, I thought I had. No wonder I'd not heard anything.

danmcd:
Last push, test-result-driven cleanup.

src/smartdc/lib/smartos_prompt_config.sh
```diff
@@ -1255,23 +1255,24 @@ your own zpool.\n"
...
 	
 	if [[ $(getanswer "skip_instructions") != "true" ]]; then
 	    printf "$message"
 	    echo "Available pre-created pools: " $(zpool list -Ho name)
 	fi
 
-	promptopt "Specify a (configured) zpool from which to boot" "none" \
-		"bootpool"
+	promptopt "Specify a (configured) zpool from which to boot" \
+		${BOOTPOOL-"none"} "bootpool"
```
jlevon, line 1269: I don't think this is right?

danmcd: It starts at "none".  If you have to loop back around because you messed up an answer, this will *remember* your prior answer (the way IP address, DNS parameters, etc. get remembered).

jlevon: TIL. Specifically:

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

When not performing substring expansion, using the form described below (e.g., ‘:-’), Bash tests for a parameter that is unset or null. Omitting the colon results in a test only for a parameter that is unset. Put another way, if the colon is included, the operator tests for both parameter’s existence and that its value is not null; if the colon is omitted, the operator tests only for existence.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+    /usr/sbin/piadm [-v | -vv] <command> [command-specific arguments]
+
+## DESCRIPTION
+
+    Historically, SmartOS booted off of a USB key or a read-only media like
+    CD-ROM.  The copy and version of the SmartOS software on one of these
+    media is called a Platform Image.  A Platform Image is described in
+    detail in the next section.
+
+    piadm(1M) manages multiple copies of Platform Images on a bootable ZFS
```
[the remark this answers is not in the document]
danmcd: Will fix.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -0,0 +1,278 @@
...
+
+        - At least one Platform Image in /POOL/boot/platform-<PI-stamp>.
+
+        - At least one Boot Image in /POOL/boot/boot-<PI-stamp>.
+
+        - A /POOL/boot/etc directory that indicates the PI-stamp for the Boot
+          Image.
+
+        - Symbolic links /POOL/boot/platform and /POOL/boot/boot that point
+          to the on-next-boot Platform Image and Boot Image.
```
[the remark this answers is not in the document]
danmcd: Will fix.

src/lib/sdc/usb-key.sh
```diff
@@ -63,6 +60,16 @@ function mount_usb_key()
...
 
+	echo "$mnt"
+}
+
+#
+# Mount the usbkey at the standard mount location (or whatever is specified).
+#
+function mount_usb_key()
+{
+	local mnt=`extract_mountpath $1`
```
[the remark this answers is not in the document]
danmcd: Didn't I do that in usb-key.sh?

```
smartos-build(~/smartos-live)[0]% pwd
/export/home/danmcd/smartos-live
smartos-build(~/smartos-live)[0]% grep '`' src/lib/sdc/usb-key.sh 
smartos-build(~/smartos-live)[1]% 
```

src/piadm.sh
```diff
@@ -0,0 +1,804 @@
...
+
+declare activestamp
+activestamp=$(uname -v | sed 's/joyent_//g')
+declare installstamp
+
+poolpresent() {
+	# Works for an empty $1, which is "all of them" or "unspecified"
+	zpool list $1 > /dev/null 2>&1
+	if [[ $? -ne 0 ]]; then
+		eecho "Pool $1 not present"
```
papertigers, line 84: Misspelled echo

papertigers: oh nvm! :)

sjorge:
:tada: one less patch to pull in when I do pi builds! congrats all, this is good stuff.

PR #951, opened 2020-09-08
OS-8207 piadm "BOOT BITS?" header is confusing

danmcd:
OS-8208 piadm man page formatting errors

### DIFF
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -94,22 +94,20 @@ piadm(1M) -- Manage SmartOS Platform Images
 
     For example:
 
-```
-[root@smartos ~]# piadm bootable
-standalone                     ==> BIOS and UEFI
-zones                          ==> non-bootable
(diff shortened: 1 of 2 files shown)


bahamat (commented):
For the man page, looks like you just removed the fences? The code blocks should be indented by eight spaces (and not `\t`) to render properly.

danmcd:
> For the man page, looks like you just removed the fences? The code blocks should be indented by eight spaces (and not `\t`) to render properly.

1.) I did just remove the fences.
2.) They aren't indented by '\t', they are indented by literally one space.  (I had emacs confirm this by uttering "M-x tabify" and literally nothing changed.)

The man rendering on the command line looks find, but the github MD-to-html does mess them up rather badly.

bahamat:
Ok, it looks fine on SmartOS, but on macOS it has some weird highlighting. Do we care?

danmcd:
I don't think we care about it being directly rendered on MacOS.

bahamat (approved):

PR #952, opened 2020-09-09
OS-8205 `piadm bootable -d` needs to be more thorough

danmcd:
Tested thus far on a SmartOS VM with an EFI-capable `zones` pool using `bootable -d` and then checking the ESP to see that it's erased.

### DIFF
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -493,7 +493,7 @@ update_boot_sectors() {
 				continue
 			fi
 			# otherwise mount the ESP and trash it.
-			tdir=`mktemp -d`
+			tdir=$(mktemp -d)
 			mount -F pcfs /dev/dsk/${a}s0 ${tdir}
 			if [[ $? -ne 0 ]]; then
 				eecho "disk $a has no PCFS ESP, it seems"
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -444,6 +444,8 @@ list() {
 update_boot_sectors() {
 	pool=$1
 	bootfs=$2
+	flag=$3
+	tdir=""
 
 	# XXX WARNING -- illumos#12894 will allow slogs.  We will need to
 	# alter the generation of boot_devices accordingly.  Generate the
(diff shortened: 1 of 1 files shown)


src/piadm.sh
```diff
@@ -486,20 +488,50 @@ update_boot_sectors() {
...
+				/${bootfs}/boot/pmbr \
+				/${bootfs}/boot/gptzfsboot \
+				/dev/rdsk/${a}${suffix} > /dev/null 2>&1 || \
+				eecho \
+				"WARNING: Can't installboot on ${a}${suffix}"
+		fi
+
 		if [[ $? -eq 0 ]]; then
 			some=1
 		fi
```
bahamat, line 528: This whole pattern seems error prone to me. `$?` will only be `0` if the *last* `umount` succeeded. What happens if all `umount` calls succeed except the last one? It doesn't seem like it's going to do the right thing.

src/piadm.sh
```diff
@@ -765,10 +788,10 @@ bootable() {
 				if [[ $? -eq 0 && \
 					-f $tdir/EFI/Boot/bootx64.efi ]]; then
 					efi="and UEFI"
-					umount $tdir
 				else
 					efi=""
 				fi
+				umount -f $tdir > /dev/null 2>&1
```
bahamat, line 794: This could be:

    umount -f "$tdir" > /dev/null 2>&1 && rmdir "$tdir"

This way, rmdir is only attempted if the `umount` succeeded, which is probably safer.

src/piadm.sh
```diff
@@ -486,20 +488,50 @@ update_boot_sectors() {
...
 		if [[ $? -eq 0 ]]; then
 			some=1
 		fi
 	done
+
+	# Partial success (altering some of the pool's disks) is good
+	# enough for command success.
 	if [[ $some -eq 0 ]]; then
-		fatal "Could not installboot(1M) on ANY vdevs of pool $2"
+		fatal "Could not make alterations on ANY vdevs of pool $2"
```
bahamat, line 534: I would just say "update" or "modify" here.

bahamat (changes requested):

src/piadm.sh
```diff
@@ -486,20 +488,50 @@ update_boot_sectors() {
...
-		# Plow through devices, even if some fail.  installboot also
-		# does loader-into-EFI-System-Partition this way.  Trailing /
-		# is important in the -b argument because boot is actually a
-		# symlink.
-		installboot -m -b /${bootfs}/boot/ /${bootfs}/boot/pmbr \
-			/${bootfs}/boot/gptzfsboot \
-			/dev/rdsk/${a}${suffix} > /dev/null 2>&1 || \
-			eecho "WARNING: Can't installboot on ${a}${suffix}"
+		if [[ "$flag" == "-d" ]]; then
+			if [[ "$suffix" == "s0" ]]; then
```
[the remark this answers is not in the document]
danmcd: Yes.  If "suffix" is s0, it's not a pool with EFI System Partition (ESP).  "suffix" is actually used for the more-common-case of installboot.  If "suffix" is s1, then s0 *is* the ESP, and we do PCFS ops on it.

src/piadm.sh
danmcd: Oh damn... I need to change that check to `if [[ $some -eq 0 && $? -eq 0 ]]`.  Thanks.

src/piadm.sh
danmcd: Thank you.

src/piadm.sh
danmcd: Okay, "modify" it is.

src/piadm.sh
danmcd: Hmmmm, upon further review, if any umount calls succeed, "some" gets set to 1.  If they fail, nothing happens ("some" remains the same).  It literally indicates at least one of the disks in the pool was successfully hit.  If it's one pool and it fails, that works.

bahamat: You know what, I missed that this is happening inside the loop, and not outside. It's fine as-is.

Sorry about that :-/

bahamat (dismissed):
lgtm

src/piadm.sh
```diff
@@ -486,20 +487,48 @@ update_boot_sectors() {
...
+				# BIOS boot, we don't care.
+				continue
+			fi
+			# otherwise mount the ESP and trash it.
+			tdir=`mktemp -d`
```
jasonbking, line 496: One small nit -- aren't we preferring $() syntax over backtics?

danmcd: Doh!  Good catch.

src/piadm.sh
```diff
@@ -486,20 +487,48 @@ update_boot_sectors() {
...
-			/dev/rdsk/${a}${suffix} > /dev/null 2>&1 || \
-			eecho "WARNING: Can't installboot on ${a}${suffix}"
+		if [[ "$flag" == "-d" ]]; then
+			if [[ "$suffix" == "s0" ]]; then
+				# BIOS boot, we don't care.
+				continue
+			fi
+			# otherwise mount the ESP and trash it.
+			tdir=$(mktemp -d)
+			mount -F pcfs /dev/dsk/${a}s0 ${tdir}
```
bahamat, line 497: Probably worth doing `${tdir:?}` here so that bash will exit just in case mktemp should happen to fail for some reason.

bahamat: We'll leave this for now, and do a file an issue to pass sellcheck.

bahamat (approved):

PR #953, opened 2020-09-12
OS-8230 want piadm.sh to be shellcheck clean

bahamat:
There *should* be no difference in behavior, except for the elimination of various undefined or bug behaviors.

That being said, there's a lot changing here so we'll want to test it against all of our known use cases.

### DIFF
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -489,9 +489,8 @@ update_boot_sectors() {
 			fi
 			# otherwise mount the ESP and trash it.
 			tdir=$(mktemp -d)
-			if mount -F pcfs "/dev/dsk/${a}s0" "${tdir:?}" ; then
-				some=1
-			else
+			if ! mount -F pcfs "/dev/dsk/${a}s0" "${tdir:?}" ; then
+				# Wrong filesystem, so skip the rest of this loop
 				eecho "disk $a has no PCFS ESP, it seems"
 				continue
 			fi
@@ -499,6 +498,7 @@ update_boot_sectors() {
 			# is using it for something ELSE also.
 			/bin/rm -rf "${tdir}/EFI"
 			umount "$tdir" && rmdir "$tdir"
+			some=1
 			# If we make it here, at least some disks had
 			# ESP and we managed to clean them out.  "some" below
 			# will get set.
@@ -597,7 +597,7 @@ remove() {
 			# Boot bits may be older than the current PI, and the
 			# current PI may not have matching boot bits for some
 			# reason. Guard against shooting yourself in the foot.
-			if ! grep -q "$pistamp" etc/version/boot; then
+			if grep -q "$pistamp" etc/version/boot; then
 				eecho "$pistamp is the current set of boot" \
 					"binaries.  Please"
 				eecho "activate another pi using" \
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -15,14 +15,15 @@
 # Copyright 2020 Joyent, Inc.
 #
 
+# shellcheck disable=1091
 . /lib/sdc/usb-key.sh
 
 eecho() {
-	echo $@ 1>&2
+	echo "$@" 1>&2
 }
 
 err() {
-	eecho $@
+	eecho "$@"
 	exit 1
 }
 
@@ -36,7 +37,7 @@ fatal() {
 }
 
 corrupt() {
-	eecho $@
+	eecho "$@"
 	exit 3
 }
 
@@ -56,7 +57,7 @@ usage() {
 vecho() {
 	if [[ $VERBOSE -eq 1 ]]; then
 		# Verbose echoes invoked by -v go to stdout, not stderr.
-		echo $@
+		echo "$@"
 	fi
 }
 
@@ -80,8 +81,7 @@ declare installstamp
 
 poolpresent() {
 	# Works for an empty $1, which is "all of them" or "unspecified"
-	zpool list $1 > /dev/null 2>&1
-	if [[ $? -ne 0 ]]; then
+	if ! zpool list "$1" > /dev/null 2>&1 ; then
 		eecho "Pool $1 not present"
 		usage
 	fi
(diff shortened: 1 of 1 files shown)


src/piadm.sh
```diff
@@ -494,15 +489,16 @@ update_boot_sectors() {
 			fi
 			# otherwise mount the ESP and trash it.
 			tdir=$(mktemp -d)
-			mount -F pcfs /dev/dsk/${a}s0 ${tdir}
-			if [[ $? -ne 0 ]]; then
+			if mount -F pcfs "/dev/dsk/${a}s0" "${tdir:?}" ; then
+				some=1
```
danmcd, line 493: Don't set "some" this early.

src/piadm.sh
```diff
@@ -494,15 +489,16 @@ update_boot_sectors() {
...
 			fi
 			# Just take out the EFI directory, in case someone
 			# is using it for something ELSE also.
-			/bin/rm -rf ${tdir}/EFI
-			umount ${tdir} && rmdir "$tdir"
+			/bin/rm -rf "${tdir}/EFI"
+			umount "$tdir" && rmdir "$tdir"
 			# If we make it here, at least some disks had
 			# ESP and we managed to clean them out.  "some" below
 			# will get set.
```
danmcd, line 504: This comment should be modified, and right after "some" should be set.

src/piadm.sh
```diff
@@ -578,44 +572,43 @@ activate() {
...
 
 		# Boot image processing.
-		if [[ -d boot-$pistamp ]]; then
+		if [[ -d "boot-$pistamp" ]]; then
 			# Boot bits may be older than the current PI, and the
 			# current PI may not have matching boot bits for some
 			# reason. Guard against shooting yourself in the foot.
-			grep -q $pistamp etc/version/boot
-			if [[ $? -eq 0 ]]; then
+			if ! grep -q "$pistamp" etc/version/boot; then
```
danmcd, line 600: This is a place where you should lose the '!'.  If the grep matches, we utter the warning and bail.

danmcd (commented):
Only found one of the wrong-bang cases you were worried about.

src/piadm.sh
bahamat: This should be good

src/piadm.sh
bahamat: I juggled this around a little bit. I think it should work for what is intended.

danmcd:
Time to pull-and-test.

src/piadm.sh
```diff
@@ -80,8 +81,7 @@ declare installstamp
 
 poolpresent() {
 	# Works for an empty $1, which is "all of them" or "unspecified"
-	zpool list $1 > /dev/null 2>&1
-	if [[ $? -ne 0 ]]; then
+	if ! zpool list "$1" > /dev/null 2>&1 ; then
```
danmcd, line 84: Would losing the quotes here be a problem?  I removed them from my local copy and my prior problems disappeared.

danmcd: Either that or you'll need to special-case an empty $1.

src/piadm.sh
```diff
@@ -197,129 +195,126 @@ mount_installmedia() {
...
+				/bin/rm -rf "${tdir:?}"
 				err "File $1 is not an ISO or a .tgz file."
 			fi
 			# We're most-likely good here.  NOTE: SmartOS/Triton
 			# PI files expand to platform-$STAMP.  Fix it here
 			# before proceeding.
-			gtar -xzf $1 -C ${tdir}/mnt
-			mv ${tdir}/mnt/platform-* ${tdir}/mnt/platform
+			gtar -xzf "$1" -C "${tdir}/mnt"
+			mv "${tdir}/mnt/platform-*" "${tdir}/mnt/platform"
```
danmcd, line 256: This breaks because I know that a freshly-extracted tarball will have platform-* in it, but I have no idea what * expands to (what stamp is in the .tgz's platform) until below.  If that breaks shellcheck, you'll need to extract the stamp from platform-* (i.e. expand * yourself).  Seems silly when shell does a great job of expansion already.  And we're guaranteed (unless it's a malicious tarball) that there's only one platform-* expansion.

src/piadm.sh
```diff
@@ -624,20 +626,17 @@ remove() {
...
-	if [[ "$currbootfs" == "${pool}/boot" ]]; then
-		output=$(zfs list -H $currbootfs 2>&1)
-		if [[ $? -eq 0 ]]; then
-			# We're bootable (at least bootable enough)
-			return 0
-		fi
+	currbootfs=$(zpool list -Ho bootfs "$pool")
+	if [[ $currbootfs == "${pool}/boot" ]]; then
+		# We're bootable (at least bootable enough)
+		zfs list -H "$currbootfs" 2>&1  && return 0
```
danmcd, line 635: Hmmm, old development had used $output.  Your change, which correctly understand "output" should no longer be used, does need to silence the output.  Perhaps:

`zfs list -H "$currbootfs" > /dev/null 2>&1 && return 0` 

?

bahamat: fixed.

src/piadm.sh
bahamat: fixed the quoting

src/piadm.sh
bahamat: special casing this for safety

src/piadm.sh
```diff
@@ -681,32 +679,27 @@ enablepool() {
 		# For now, proceed clobber-style.
 	fi
 
-	output=$(zfs list -H $bootfs 2>&1)
-	if [[ $? -ne 0 ]]; then
+	if ! zfs list -H "$bootfs" 2>&1; then
```
danmcd, line 682: Same thing here:  the output needs to go to /dev/null.

bahamat: Fixed

danmcd (approved):
Bug seem to have been shaken out.

danmcd:
Testing notes are in the ticket.

PR #956, opened 2020-09-17
OS-8206 Add "ipxe" as source for piadm(1M) on Triton CNs

danmcd:


danmcd:
Very much a WIP.

src/piadm.sh
```diff
@@ -199,6 +218,104 @@ mount_installmedia() {
...
+	/bin/rm -rf ./platform-ipxe ./boot-ipxe ./etc/version/ipxe
+
+	if [[ $bootstamp == "ipxe" ]]; then
+		# Remove the symlinks.
+		# iPXE CAN NOT have mismatched boot and platform symlinks.
+		# It is safe to remove boot as well.
+		rm ./platform ./boot
+		vecho "Removal of active iPXE leaves nothing for $bootfs to find."
+		vecho "Use 'piadm install' to provide something."
+		vecho "If you're using 'piadm update', that will happen now."
```
bahamat, line 313: This could also note that external boot media will be required to boot

src/piadm.sh
```diff
@@ -199,6 +218,104 @@ mount_installmedia() {
...
+	mkdir boot-ipxe
+	tar -cf - -C ${TRITON_IPXE_BOOT} . | tar -xf - -C boot-ipxe
+
+	# Populate loader.conf.
+	# XXX KEBE SAYS make loader.conf work for ipxe.
+	# XXX KEBE SAYS FILL ME IN!  AND ONCE FILLED IN CHECK ME!
+	sed 's/headnode="true"/headnode="false"/g' \
+	    < boot-ipxe/loader.conf.tmpl > boot-ipxe/loader.conf
+	echo 'ipxe="true"' >> boot-ipxe/loader.conf
+	echo 'smt_enabled="true"' >> boot-ipxe/loader.conf
```
bahamat, line 277: Looks like there's still more you plan to do here, but my initial thought is that perhaps we should leave set_enabled at whatever value it currently is. We don't want to flip this if it's currently set opposite.

src/piadm.sh
```diff
@@ -199,6 +218,104 @@ mount_installmedia() {
...
+# from the platform directory, it will instead boot "ipxe.lkrn" and
+# "default.ipxe" from the boot directory.
+#
+# Because of the desire to have a Triton Compute Node possibly boot to
+# a recovery SmartOS, we will never install iPXE on the ESP.
+#
+# "bootfs" has been set already.
+install_ipxe() {
+	cd /${bootfs}
+	if [[ -e platform-ipxe || -e boot-ipxe ]]; then
```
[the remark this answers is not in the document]
danmcd: They probably should.

NOTE:  upcoming changes go beyond just your code review comments.  :)

src/piadm.sh
```diff
@@ -199,6 +218,104 @@ mount_installmedia() {
...
+	cat_readme > ./platform-ipxe/README
+	echo "ipxe" > ./platform-ipxe/etc/version/platform
+
+	# Entry to this function is predicated by existence of
+	# TRITON_IPXE_LKRN and TRITON_IPXE_ARCHIVE.  Assume
+	# TRITON_IPXE_ETC is there too.
+	mkdir -p etc/version
+	cp -f ${TRITON_IPXE_ETC}/version/ipxe etc/version/.
+	# So 'piadm list' doesn't seize up...
+	echo "ipxe" > etc/version/boot
```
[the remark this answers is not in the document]
danmcd: Hmmm... right now, that will make piadm(1M) (especially OLDER versions) fail a reality check.  The design idea is that "ipxe" is a special-case-name for a timestamp/boot-stamp.

I can make "boot" match what's there, but it'll require more work in "piadm list".  Mightn't be able to get to it tonight, however.  I've marked the possible trouble spots with more KEBE comments.

src/piadm.sh
```diff
@@ -199,6 +218,104 @@ mount_installmedia() {
...
+	vecho "installing ipxe version: " $(cat etc/version/ipxe)
+
+	mkdir boot-ipxe
+	tar -cf - -C ${TRITON_IPXE_BOOT} . | tar -xf - -C boot-ipxe
```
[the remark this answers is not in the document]
danmcd: Am I guaranteed to have that?  I thought that was an "extra".

src/piadm.sh
danmcd: Correct, all of this is very much up in the air still (hence the KEBE comments).  My machines seem to have it set to enabled, but what do I know?

src/piadm.sh
danmcd: Actually rewhacked this whole section to make "ipxe" more like any other PI and not something super-special (modulo "update").

danmcd (commented):
Update incoming...

src/piadm.sh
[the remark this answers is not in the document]
bahamat: Yes. It's in the platform image.

src/piadm.sh
bahamat: I think the default was changed at one point, but a triton install that was set up before the default change will retain `smt_enabled=true`. It's my understanding that new triton installs will default to `false`.

But in either case, if a user has explicitly chosen a value overriding the default, we want to retain that choice and not potentially change it on them unexpectedly.

src/piadm.sh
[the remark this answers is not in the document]
danmcd: Hmmm, I used tar also because the normal PI image installs use it (tar from CD, USB, or downloaded .iso, untar in /$bootfs), but you're most likely right that rsync will work better here.  Also on the todo list now.

src/piadm.sh
danmcd: OS-8209 may help here.  Jorge S. is working it right now, see #958 .

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -5,14 +5,15 @@ piadm(1M) -- Manage SmartOS Platform Images
...
-    piadm activate|assign <PI-stamp> [ZFS-pool-name]
+    piadm activate|assign <PI-stamp/ipxe> [ZFS-pool-name]
     piadm avail
     piadm bootable
     piadm bootable [-dr] <ZFS-pool-name>
     piadm bootable -e [ -i <source> ] <ZFS-pool-name>
     piadm install <source> [ZFS-pool-name]
     piadm list [ZFS-pool-name]
     piadm remove <PI-stamp> [ZFS-pool-name]
+    piadm update ipxe [ZFS-pool-name]
```
danmcd, line 16: Reviewers: I'm contemplating just making this a single-keyword `update-ipxe`.

src/piadm.sh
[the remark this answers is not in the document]
danmcd: Actually, I have one question:  at this point in the code, we are guaranteed to have ./boot-ipxe not exists.  Is rsync still *better* than tar-to-tar in this situation?  I'm used to using rsync for remote ops, not local, so your experience will inform me here, @bahamat

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -69,6 +70,15 @@ piadm(1M) -- Manage SmartOS Platform Images
...
     changes also that need to accompany it.
 
+## iPXE
+
+    On Triton Compute Nodes, which run SmartOS booted off a network, stored
+    locally is an implementation of iPXE, an enhanced form of Preboot
+    eXecution Environment (PXE). piadm(1M) can install iPXE in lieu of the
+    normal Boot Image and Platform Image.  The word "ipxe" can be used in
+    place of a PI stamp or a source.  Unlike Platform Images, ipxe can only
+    have one instance maintained on a given bootable pool.
```
bahamat, line 80: For clarity, I think it should also state that the version of ipxe is controlled by the Triton headnode, and piadm will simply put the provided ipxe in place, and possibly refer to the sdcadm man page.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -120,14 +130,16 @@ piadm(1M) -- Manage SmartOS Platform Images
 
     piadm(1M) commands and options are:
 
-      piadm activate <PI-stamp> [ZFS-pool-name]
-      piadm assign <PI-stamp> [ZFS-pool-name]
+      piadm activate <PI-stamp/ipxe> [ZFS-pool-name]
+      piadm assign <PI-stamp/ipxe> [ZFS-pool-name]
```
bahamat, line 134: As above, prefer pipe.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -120,14 +130,16 @@ piadm(1M) -- Manage SmartOS Platform Images
...
 
         Activate a Platform Image for the next boot, on a specified ZFS pool
         if there are more than one bootable pools imported.  It is up to the
         administrator to know which pool the system will actually boot.  If a
         boot image with the specified PI-stamp is unavailable, a warning will
-        be issued but the new PI will be activated anyway.
+        be issued but the new PI will be activated anyway.  If "ipxe" is
+        specified, the pool will boot the system into iPXE instead of
```
bahamat, line 141: Should this be something like "loader will default to ipxe first, then fall back to the selected platform image if there is no response"?

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -183,18 +196,28 @@ piadm(1M) -- Manage SmartOS Platform Images
...
+            use the update subcommand.  Also, an installation of "ipxe" will
+            override any existing Boot Images and Platform Images.
 
       piadm list [ZFS-pool-name]
 
         Lists the available platform images (and boot images) on bootable
         pools.
 
-      piadm remove <PI-stamp> [ZFS-pool-name]
+      piadm remove <PI-stamp/ipxe> [ZFS-pool-name]
```
bahamat, line 210: As above, prefer pipe.

src/piadm.sh
```diff
@@ -37,20 +37,21 @@ fatal() {
...
 	eecho "Usage: piadm [-v] <command> [command-specific arguments]"
 	eecho ""
-	eecho "    piadm activate|assign <PI-stamp> [ZFS-pool-name]"
+	eecho "    piadm activate|assign <PI-stamp/ipxe> [ZFS-pool-name]"
 	eecho "    piadm avail"
 	eecho "    piadm bootable [-d] [-e [-i <source>]] [-r] [ZFS-pool-name]"
 	eecho "    piadm install <source> [ZFS-pool-name]"
 	eecho "    piadm list <-H> [ZFS-pool-name]"
-	eecho "    piadm remove <PI-stamp> [ZFS-pool-name]"
+	eecho "    piadm remove <PI-stamp/ipxe> [ZFS-pool-name]"
```
bahamat, line 53: As above, prefer pipe.

bahamat (changes requested):

bahamat:
I did not mean to click that button. Sorry.

man/usr/share/man/man1m/piadm.1m.md
bahamat: It occurs to me that we maybe shouldn't have verbs for ipxe at all. Instead, piadm would be used to specify the "fallback" PI only, but the boot bits indiscriminately use the loader/ipxe provided by the HN that are already on disk. The more I think about it, I don't think we're ready to have CNs booting a local PI for normal operation.

We may also want Triton CNs to only be able to install platform images that are available from the headnode, and not look in manta. This is complicated by the fact that the HN doesn't retain the PI tarball after unpacking. Maybe this could be fixed by having sdcadm platform add the platform images to imgapi, then Triton CNs could fetch it from imgapi. By default, CNs won't have a default gateway to reach manta anyway. So the process for Triton would be something like this, executed from the headnode:

```
curl -o <some image>
sdc-oneachnode -c -g <image> -d /var/tmp
sdc-oneachnode -c 'piadm bootable -e /var/tmp/<image> zones'
```

bahamat: My previous comments about "prefer pipe" are kind of moot if we go this route.

src/piadm.sh
```diff
@@ -168,6 +179,14 @@ avail() {
 	# Don't print ones old enough to NOT contain piadm(1M) itself.
 	"${CURL[@]}" "${URL_PREFIX}/?limit=1000" | json -ga -c \
 		"this.name.match(/Z$/) && this.name>=\"$activestamp\"" name
+
+	#
+	# Additionally, check to see if we have a Triton CN, where we can
+	# create an ipxe boot. Restrict to searching only locally on-disk.
+	#
+	if [[ -f "$TRITON_IPXE_LKRN" && -f "$TRITON_IPXE_ARCHIVE" ]]; then
```
bahamat, line 187: I think we should also include checking bootparams to ensure that neither `headnode=true` nor `smartos=true` are set. I.e., we don't want you to be able to accidentally set your headnode to boot ipxe, because it may never come back.

bahamat: And thinking about this again, we'll definitely want `piadm update ipxe` to move new bits into place from the (presumably) newer local copy.

But, I'm thinking `piadm bootable -i <image> <pool>`, on a Triton node uses the ipxe that it already has, and the specified image will use only the PI.

danmcd: NOTE:  `bootable -i` is now disallowed on Triton CNs.  For the backup PI you get either what you're running, or the "default" PI.

danmcd:
Test run from this morning using the latest piadm per this PR:

```
[root@larry (kebecloud) ~]# /tmp/piadm list
PI STAMP               BOOTABLE FILESYSTEM            BOOT IMAGE NOW  NEXT 
20200930T173623Z       zones/boot                     next       no   no  
ipxe                   zones/boot                     next       no   yes 
[root@larry (kebecloud) ~]# /tmp/piadm -v list
PI STAMP               BOOTABLE FILESYSTEM            BOOT IMAGE NOW  NEXT 
20200930T173623Z       zones/boot                     next       no   no  
ipxe(20200626T133005Z) zones/boot                     next       no   yes 
[root@larry (kebecloud) ~]# zfs list -t all | grep boot
zones/boot                                         600M   284G   304M  /zones/boot
zones/boot@baseline                                297M      -   300M  -
[root@larry (kebecloud) ~]# zfs rollback zones/boot@baseline
[root@larry (kebecloud) ~]# /tmp/piadm -v list
PI STAMP               BOOTABLE FILESYSTEM            BOOT IMAGE NOW  NEXT 
20200910T155139Z       zones/boot                     none       no   no  
ipxe(20200219T112943Z) zones/boot                     next       no   yes 
[root@larry (kebecloud) ~]# 
[root@larry (kebecloud) ~]# /tmp/piadm -v bootable
zones                          ==> BIOS 
[root@larry (kebecloud) ~]# /tmp/piadm -v bootable -e
To enable a pool for booting, please specify at last a pool

Usage: piadm [-v] <command> [command-specific arguments]

    piadm activate|assign <PI-stamp> [ZFS-pool-name]
    piadm avail
    piadm bootable [-d] [-e [-i <source>]] [-r] [ZFS-pool-name]
    piadm install <source> [ZFS-pool-name]
    piadm list <-H> [ZFS-pool-name]
    piadm remove <PI-stamp> [ZFS-pool-name]
    piadm update [ZFS-pool-name]

[root@larry (kebecloud) ~]# /tmp/piadm -v bootable -e zones
Pool zones appears to be bootable.
Use 'piadm update' to update the CN's iPXE and backup PI.
[root@larry (kebecloud) ~]# /tmp/piadm -v update
Selecting lone boot pool zones by default.
Updating backup PI to 20201002T173911Z
Installing as a backup Platform image PI stamp 20201002T173911Z
making platform-20201002T173911Z directories
Pulling unix
Pulling boot_archive
Pulling boot_archive.hash
Pulling boot_archive.manifest
Pulling boot_archive.gitstatus
...success installing 20201002T173911Z
Updating iPXE provided by headnode (ver: 20200626T133005Z)
Updating boot provided by headnode (ver: 20200930T173623Z)
[root@larry (kebecloud) ~]# /tmp/piadm -v update
Selecting lone boot pool zones by default.
No updates needed for iPXE and its loader.
If you think there should be an update, run
'sdcadm experimental update-gz-tools' on your Triton Head Node
[root@larry (kebecloud) ~]# /tmp/piadm update
[root@larry (kebecloud) ~]# echo $?
0
[root@larry (kebecloud) ~]# /tmp/piadm -v list
PI STAMP               BOOTABLE FILESYSTEM            BOOT IMAGE NOW  NEXT 
20201002T173911Z       zones/boot                     none       yes  no  
ipxe(20200626T133005Z) zones/boot                     next       no   yes 
[root@larry (kebecloud) ~]# /tmp/piadm list
PI STAMP               BOOTABLE FILESYSTEM            BOOT IMAGE NOW  NEXT 
20201002T173911Z       zones/boot                     none       yes  no  
ipxe                   zones/boot                     next       no   yes 
[root@larry (kebecloud) ~]# 
```

PR #957, opened 2020-09-18
OS-8232 Switch to smartos-build-tools package

jperkin:


jperkin:
I've tested this in two configurations, firstly with a fresh zone testing that the package installs correctly and successfully builds smartos-live, and secondly by manually installing the legacy broken packages on the zone to test that the `pkg_delete` section works correctly.

timfoster (approved):
This change looks correct to me. Thanks for sorting it out!

danmcd (approved):
Please copy the testing paragraph from this ticket to Jira, for redundancy & following what we like to do in illumos.

PR #958, opened 2020-09-18
OS-8209 piadm should allow injecting loader.*.local

sjorge:
After this change piadm will try to symlink `loader.conf.local` and `loader.rc.local` on the install action.

The symlinking is only done when:
- installing boot bits
- either loader.conf.local or loader.rc.local exist in /BOOTFS/common/

This should make it easier for people to load for example a custom ppt_matches and ppt_aliases file.

Testing I've done:
- build a new pi with this applied
- install pi and boot it
- create /standalone/boot/common
- copy my ppt_aliases, ppt_matches, sd.conf, ... to the common directory (mirroring the final location)
- copy my existing loader.conf.local (and modified the paths) to common directory
- create test loader.rc.local that swaps bg and foreground colors
- build a new pi
- install new pi using `piadm install XXX.iso`
- verified that a relative symlink was present in boot-STAMP
- reboot and double check serial output for my files to be loaded (OMG the VGA output is white and I regret the choice)
- remove dummy loader.rc.local (leaving broken symlink)
- reboot
- loader.conf.local still loaded but back to normal colors on VGA (hurray)

Down side is that if you remove either loader.conf.local or loader.rc.local from your common directory it leaves a broken symlink.  It's probably safe to assume if the sysadmin is messing with this, they are smart enough to clean up the symlinks.
Even if they don't, there are no errors when booting.

### DIFF
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -69,6 +69,11 @@ piadm(1M) -- Manage SmartOS Platform Images
     itself without issue.  Occasionally, however, a PI will have Boot Image
     changes also that need to accompany it.
 
+    Boot images can be customized by providing loader.conf.local and/or
+    loader.rc.local files in POOL/boot/common/. When a new boot image is
+    installed using `piadm install` these files are symlinked into the
+    images.
+
 ## BOOTABLE POOLS
 
     A SmartOS bootable pool (POOL in the examples) contains:
@@ -293,3 +298,8 @@ The following exit values are returned:
     if it's `zones`, a machine can still be booted with a USB stick, CD-ROM,
     or other method of booting SmartOS.  A bootable pool can then be
     repaired using piadm(1M) from the USB stick or CD-ROM.
+
+    Boot image customization is done at install time, other boot images
+    already present will not be updated. Older versions of piadm can still
+    be used to switch to and from boot images that have been customized but
+    won't be able to customize images.
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -72,7 +72,7 @@ piadm(1M) -- Manage SmartOS Platform Images
 
     The behavior of loader can be controlled by providing loader.conf.local
     and/or loader.rc.local files in the ${BOOTPOOL}/boot-${VERSION}
-    directory. Loader config files can also be placed in ${BOOTPOOL}/common,
+    directory. Loader config files can also be placed in ${BOOTPOOL}/custom,
     and will be used by all subsequently installed boot images.
 
     See loader.conf(4) and loader(5) for the format of these files.
--- a/man/usr/share/man/man1m/piadm.1m.md
+++ b/man/usr/share/man/man1m/piadm.1m.md
@@ -250,6 +250,49 @@ piadm(1M) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+### Customizing a boot image to load /etc/ppt_aliasses and /etc/ppt_matches
+
+```
+ [root@smartos ~]# piadm list
+ PI STAMP           BOOTABLE FILESYSTEM            BOOT IMAGE   NOW   NEXT
+ 20200714T195617Z   standalone/boot                next         yes   yes
+ [root@smartos ~]# mkdir -p /standalone/boot/common
+ [root@smartos ~]# cat > /standalone/boot/common/loader.conf.local <<EOF
+ > ppt_aliases_type="file"
+ > ppt_aliases_name="/common/etc/ppt_aliases"
+ > ppt_aliases_flags="name=/etc/ppt_aliases"
+ > ppt_matches_load="YES"
+ > ppt_matches_type="file"
+ > ppt_matches_name="/common/etc/ppt_matches"
+ > ppt_matches_flags="name=/etc/ppt_matches"
+ > EOF
+ [root@smartos ~]# mkdir -p /standalone/boot/common/etc
+ [root@smartos ~]# cat > /standalone/boot/common/etc/ppt_aliases <<EOF
+ > ppt "/pci@5e,0/pci8086,2030@0/pci8086,37c0@0/pci8086,37c5@3/pci15d9,37d2@0"
+ > EOF
+ [root@smartos ~]# cat > /standalone/boot/common/etc/ppt_matches <<EOF
+ > pciex8086,37d2
+ > EOF
+ [root@smartos ~]# piadm install 20200910T013122Z
+ [root@smartos ~]# ls -l /standalone/boot/boot-20200910T013122Z
+ total 527245
+ -r--r--r--   1 root     root        1249 Sep 10 04:29 cdboot
+ drwxr-xr-x   2 root     root           3 Sep 10 04:29 defaults
+ -rw-------   1 root     root     268435456 Sep 10 04:29 efiboot.img
+ drwxr-xr-x   2 root     root          22 Sep 10 04:29 forth
+ -r--r--r--   1 root     root      142848 Sep 10 04:29 gptzfsboot
+ -r--r--r--   1 root     root       12303 Sep 10 04:29 joyent.png
+ -r--r--r--   1 root     root      405504 Sep 10 04:29 loader
+ -rw-r--r--   1 root     root         291 Sep 10 04:29 loader.conf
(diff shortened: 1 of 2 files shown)


sjorge:
I think there was an internal ticket for allowing something like this, but I can't find the OS-XXX number for it.

sjorge:
Found it, it was OS-8209

danmcd:
Okay, this is good.  I think we'll need some man page text to go along with it, however.

And unrelated, but along these lines, we need to see how painful OS-8231 is (not only will we need install() help like this one, but we'll need remove() help as well).

I'd like to see the accompanying man page changes as well, please.

danmcd:
One other corner case:  What about systems with already-deployed PIs?  We will need to flesh out how to catch these ones up or document that you must uninstall/reinstall.

sjorge:
Updated the man page with a some info, an example and a note about already installed images.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -245,6 +250,49 @@ piadm(1M) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+### Customizing a boot image to load /etc/ppt_aliasses and /etc/ppt_matches
+
+```
+ [root@smartos ~]# piadm list
+ PI STAMP           BOOTABLE FILESYSTEM            BOOT IMAGE   NOW   NEXT
+ 20200714T195617Z   standalone/boot                next         yes   yes
+ [root@smartos ~]# mkdir -p /standalone/boot/common
```
danmcd, line 259: I'm gonna stop you right there.  I wonder if we should be doing this as part of `piadm bootable -e` or `piadm bootable -r`.  Let's not push that on the user, per se (but yes, you've documented it well above).

src/piadm.sh
```diff
@@ -359,6 +359,13 @@ install() {
 		tar -cf - -C "${tdir}/mnt/boot" . | \
```
danmcd, line 359: As I said in the man page, I wonder if we should make the "common" directory as part of pool refresh or poll enabling.  Basically the refresh_or_disable_pool() function and the enablepool() function should have the ability to mkdir /$bootfs/common (and leave it to the user).

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -69,6 +69,11 @@ piadm(1M) -- Manage SmartOS Platform Images
...
     changes also that need to accompany it.
 
+    Boot images can be customized by providing loader.conf.local and/or
+    loader.rc.local files in POOL/boot/common/. When a new boot image is
+    installed using `piadm install` these files are symlinked into the
+    images.
+
 ## BOOTABLE POOLS
 
     A SmartOS bootable pool (POOL in the examples) contains:
```
danmcd, line 79: Toward the bottom of this list, you should add a common/ directory (and mention that -r can create it if it's not there already), as well as document that the two *.local files will be symlinked to by all boot images.  (It may involve moving your new text above, OR repeating yourself, to make the point.)

danmcd (commented):
Good progress. The "create common/" occurred to me as I was reading your example.

danmcd:
Even if it's empty, it should be there, as it serves a documented purpose.  For the ipxe off-disk boot I'm doing (OS-8206) the "platform-ipxe" directory needs to be there even if it's mostly a NOP.  It helps give guidance to the new user (and as I think about it, a README should be there as well).

sjorge:
Hmmm if we are going to include README's there must be a better way to create them instead of having piadm do it.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -69,6 +69,11 @@ piadm(1M) -- Manage SmartOS Platform Images
     itself without issue.  Occasionally, however, a PI will have Boot Image
     changes also that need to accompany it.
 
+    Boot images can be customized by providing loader.conf.local and/or
+    loader.rc.local files in POOL/boot/common/. When a new boot image is
+    installed using `piadm install` these files are symlinked into the
+    images.
```
bahamat, line 75: To make this more clear, I think this paragraph should be changed to read as follows:

> The behavior of loader can be controlled by providing loader.conf.local and/or loader.rc.local files in the ${BOOTPOOL}/boot-${VERSION} directory. Loader config files can also be placed in ${BOOTPOOL}/common, and will be used by all subsequently installed boot images. See loader.conf(4) and loader(5) for the format of these files.

man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -293,3 +341,10 @@ The following exit values are returned:
     if it's `zones`, a machine can still be booted with a USB stick, CD-ROM,
     or other method of booting SmartOS.  A bootable pool can then be
     repaired using piadm(1M) from the USB stick or CD-ROM.
+
+    Boot image customization is done at install time, other boot images
+    already present will not be updated. Older versions of piadm can still
+    be used to switch to and from boot images that have been customized but
+    won't be able to customize images. Because symlinks are used, changes
+    to common/loader.conf.local and common/loader.rc.local will be
+    be applied to all boot images that have been customized.
```
bahamat, line 350: If the paragraph change I suggested for hunk 1 of this file, I don't think this paragraph will be necessary.

man/usr/share/man/man1m/piadm.1m.md
bahamat: I don't really like this section and I think it should be entirely removed.

These config files are how to control the behavior of *loader*. While loading modules is supported, it's not intended to be a general use tool kit for overriding portions of the platform image.

The example of modifying these particular files, IMO is overly specific, and in general is something that should be avoided unless you *really* know what you're doing. There are a very few users who use this frequently, but the vast majority of users will have no need. Putting this right in the man page here, to me, makes it seem like this is a common, or recommended practice, that most users will need. When in reality almost all users should not be modifying these files.

bahamat (changes requested):

man/usr/share/man/man1m/piadm.1m.md
sjorge: I agree with the KISS approach, I will switch to the paragraph suggested here https://github.com/joyent/smartos-live/pull/958#discussion_r495319106.

That just leaves open if we should create `common/` on bootable and if we should update all existing images or not, I'm leaning on maybe on the first and no for the later.

sjorge:
I've gone with @bahamat 's reworked text, which does indeed make the 2nd block unnecessary. I've also dropped the example.

I think the open issues now are :
- [ ] do we create `common/` on `piadm bootable -e` and `piadm bootable -r` 
- [ ] do we update existing images?

Personally I'm leaning towards no on updating existing images, original and revised man page also state that this is the behavior.

I'm neutral to creating `common/` automatically, I see no use in having an empty directory around. Including a README would be useful if we do create one, but how do we ship it? Do we embed it in piadm.sh? (seems like a silly idea) or do we simply add a blurb that says `see man 1 piadm` ?

bahamat:
I would say no on both counts.

* We don't auto create /opt/custom, no need to create $pool/common
* People can easily symlink it to older platforms if they want to. IMO, this is more about not having to repeat myself every time I install a new image.

But...should it be $pool/custom to be consistent with /opt/custom? I would like to bias for consistency so my preference would be to change it to custom rather than common if there's no objection from anyone.

sjorge:
I originally had it as custom but common was mentioned on IRC. I'll update it tonight and rename the things and do another build.

sjorge:
Done and still works

1. build new image
2. piadm install image
3. boot (still loads from common, as piadm was still the old one)
4. piadm install latest
5. verify symlinks
6. (oops) they are missing, which after debugging is to be expected! there is no custom/ dir
7. remove latest pi
8. mv common -> custom and update my loader.conf.local
9. piadm install latest
10. symlink is now present

I also checked that `man piadm` mentions the new dir.

danmcd:
I hadn't realized the prior-art of /opt/custom, so /$BOOTFS/custom makes far more sense -- thank you both.  Going to look over the code & man changes to see what else pops out at me.

danmcd:
This is much simpler and I hope addresses Brian's concern.

sjorge:
Keeping it simple is probably best, just symlinking thing in on install should generally all that is needed.
What the sysadmin does, is not really a concern for us.

Personally I load up ppt_aliases and ppt_matches, but others have different needs and are free to do so.

bahamat (approved):

bahamat:
This looks great. Thanks for doing that work!

danmcd:
I am going to copy the most recent testing notes to the ticket.

danmcd:
Okay, pushing this.

PR #959, opened 2020-09-18
OS-8234 illumos-joyent should use jdk11 for builds

jasonbking:


danmcd:
I'm wondering if we could jump right to openjdk11?

jperkin:
In terms of package availability, yes, openjdk11-11.0.2 is available in 2018Q4 onwards.

danmcd (approved):
Ship it!

jperkin (approved):
Yeh assuming the build looks good it's a +1 from me too, we build openjdk11 out of our custom repository so should be very future-proof from a packaging perspective.

PR #962, opened 2020-10-20
OS-8240 SmartOS interactive installation needs to allow non-media ins…

danmcd:
…tallation


danmcd:
Tested with alternate-source on Packet.

bahamat (approved):
LGTM

PR #965, opened 2020-12-01
OS-8250 Create 16gb.img in smartos-live

danmcd:
Subject says it all.


danmcd:
NOTE:  This is Pedro's work, I'm pushing it along (no pun intended).

kusor (approved):
Thanks @danmcd!

I really dunno why we don't build these images with sdc-headnode. I'll make a note there regarding where to go when adding a new image size is required due to whatever the reason.

papertigers (approved):

PR #968, opened 2021-01-15
OS-8261 piadm(1M) shouldn't case-correct already-correct os/ entries

danmcd:


src/piadm.sh
```diff
@@ -1015,7 +1015,11 @@ bringup_HN() {
 	vecho "Case-correcting os/ entries."
 	cd ./os
```
bahamat, line 1016: I know this isn't related to the change you're making, but this should be pushd, with an associated popd before the end of the function.

As it is, this function `cd`'s into os, and then the `cwd` for the whole program is different after calling the function.

bahamat (approved):
Approving this, because this specific change is ok, and my other comment about the `cd` is irrelevant to this specific change.

I'm ok with either:

* a separate bug/change for cd
* fixing cd in this PR (in which case, I'll re-approve)

src/piadm.sh
danmcd: For this case, it really doesn't matter, because bringup_HN() is called near the end of enablepool(), only followed by an update_boot_sectors(), which doesn't use any PWD-relative accesses.  Once enablepool() is done, bootable() returns, and the command is finished.

I can either comment to that effect, OR do your pushd/popd.  As I think about it, I honestly think `cd ..` after the for loop is better.  (pushd/popd are noisy by default)

bahamat (approved):

PR #970, opened 2021-02-03
OS-8265: TRITON-2137 has broken manifest for qlocker

danmcd:
Just detected now, over 6mos later.


bahamat (approved):

papertigers:
Oops!

papertigers (approved):

jlevon:
Oops, indeed, sorry about that

danmcd:
> Oops, indeed, sorry about that

The surprising part is that it took this long to actually bite us.  Would ask for an IA, but I don't think you can.  :)

PR #971, opened 2021-02-10
Workaround for OS-7666: vminfod times out while starting with thousands of snapshots

plitc:


twhiteman:
Thanks - these changes look like they are worthwhile, but I must note that I don't think any of these changes would affect vminfod:
- `imgadm` is not used by vminfod (as far as I could see)
- `VM.js` changes affect vmadm - notably to the provision, reprovision and archiveZone methods
- `sysinfo` changes would affect a non-global zone (i.e. when sysinfo is called from inside of a zone)
- others are utility commands and/or used for tests

plitc:
what about **src/vm/node_modules/vmload/vmload-datasets.js** ?

twhiteman:
> what about src/vm/node_modules/vmload/vmload-datasets.js ?

Yes - all of the main code for vminfod will be inside these two directories:
- src/vm/node_modules/vminfod/
- src/vm/node_modules/vmload/  (shared vmadm and vminfod code)

twhiteman (approved):

PR #974, opened 2021-02-18
OS-8271 piadm(1M) needs Head Node text

danmcd:


man/usr/share/man/man1m/piadm.1m.md
```diff
@@ -200,6 +210,18 @@ piadm(1M) -- Manage SmartOS Platform Images
...
 
+	For Triton Head Nodes, the -i option is also disabled.  When invoked
+	with -e on a Head Node, the piadm(1M) command will attempt to convert
+	a pool to be bootable for a Triton Head Node.  If a Head Node is
+	booting from a USB key, the boot data comes from the USB Key.  If
+	a Head Node is booting from another pool, the boot data comes from
+	the current booted pool.  After invoking `piadm bootable -e $POOL`,
+	$POOL can boot the Triton Head Node, BUT any pre-reboot USB-key
+	operations (even if the USB key is actually the bootable ZFS
+	filesystem), will not copy over to the newly-enabled bootable pool.
```
bahamat, line 221: Does this always apply to the previous boot media in any case? Or only when changing from a USB?

The way it reads here, is that if you’re using poolA, and change to poolB, you don’t need to reboot and it’ll always do the right thing.

danmcd: It does apply to any previous boot media.  And good point.  Lemme reword.

bahamat (approved):

PR #978, opened 2021-03-03
OS-8274 smartos jenkins labels should select new build machines

timfoster:


Jenkinsfile
```diff
@@ -5,7 +5,7 @@
  */
 
 /*
- * Copyright 2020 Joyent, Inc.
+ * Copyright 2021 Joyent, Inc.
  */
 
 @Library('jenkins-joylib@v1.0.5') _
```
danmcd, line 11: Hmmm, why was it bumped prior?

danmcd (approved):
Ship it, but I won't IA this until I get an answer about the jenkins-joylib version.

Jenkinsfile
danmcd: Looks like for TOOLS-2454.

I take it since jenkins-joylib didn't change this time we can merely catch up all of the other Jenkinsfiles to be the new version?  (I looked at jenkins-joylib and didn't see any versioning changes so I'm not sure how those version bumps work...)

timfoster: Good question. This was because smartos-live was actually the reason for implementing the new code that arrived in the 1.0.5 version of jenkins-joylib, and so became the first repository to use it.

Specifically, because smartos-live is multi-stage pipeline, we wanted to give ~os users insight into which stage of each pipeline had completed in the notification messages. This was done by adding a `comment` parameter to each `joyMattermostNotification()` call. For example, here's an excerpt of smartos-live's `Jenkinsfile` showing the gcc 4 pipeline notification:

```
                    joyMattermostNotification(
                        channel: 'os', comment: 'gcc4')
```

The "1.0.5" version string used here points to a git tag on the `jenkins-joylib.git` repository, which causes Jenkins to then pull in that exact version of the groovy scripts, which extend Jenkinsfiles. (the library provides support for common labels across Manta/Triton builds, Mattermost notifications, and an easy way to invoke the `make all release publish buildimage bits upload` Makefile targets, common to most Manta/Triton repos)

timfoster: Just realized I didn't answer your question,

> I take it since jenkins-joylib didn't change this time we can merely catch up all of the other Jenkinsfiles to be the new version?

yes, that's exactly it. That was only a single-repo push, and since the other repos didn't use the new functionality anyway, there didn't seem to be any point. We're only changing it now, since we're already in those Jenkinsfiles.

PR #987, opened 2021-05-03
OS-8284 want os zones to be able to mount volumes from other zones

bahamat:


src/vm/node_modules/VM.js
```diff
@@ -12064,8 +12067,6 @@ function deleteZone(uuid, log, callback)
                 'VM has local data volumes, checking other containers for '
                 + '--volumes-from');
 
-            // TODO can we safely limit to the same owner_uuid?
```
bahamat, line 12067: I removed this line because I don't think we can ever reasonably limit this to owner_uuid. Standalone SmartOS often will not have an owner_uuid and we still need to check for it there.

src/vm/node_modules/VM.js
```diff
@@ -12055,16 +12067,14 @@ function deleteZone(uuid, log, callback)
...
                 cb();
                 return;
             }
 
             log.info({uuid: vmobj.uuid, volumes: data_volumes},
                 'VM has local data volumes, checking other containers for '
-                + '--volumes-from');
-
-            // TODO can we safely limit to the same owner_uuid?
+                + 'dependant filesystems');
```
danmcd, line 12077: s/dependant/dependent/g

And I mean "/g" there, I see other identical (copy/pasted?) misspellings.

src/vm/node_modules/VM.js
```diff
@@ -8688,11 +8697,11 @@ function installZone(payload, log, callback)
             // Create any filesystems now that have the 'create' flag set.
             // Note that currently if you reprovision, all these created volumes
             // will be destroyed and recreated as well since they're children of
-            // the zoneroot. This only currently works with "docker" VMs.
+            // the zoneroot. This does not work with HVMs.
             if (receiving
                 || reprovisioning
                 || !payload.hasOwnProperty('add_filesystems')
-                || !vmobj.docker) {
+                || vmobj.hvm) {
```
danmcd, line 8704: Per MM, test on LX.  If it works, great.  If it doesn't, add a `|| vmobj.lx` or whatever-it-is check.

danmcd (commented):
Almost there...

src/vm/node_modules/VM.js
bahamat: Fixed.

src/vm/node_modules/VM.js
bahamat: Notes added to jira.

danmcd (approved):

src/vm/node_modules/VM.js
```diff
@@ -8688,11 +8697,11 @@ function installZone(payload, log, callback)
             // Create any filesystems now that have the 'create' flag set.
             // Note that currently if you reprovision, all these created volumes
             // will be destroyed and recreated as well since they're children of
-            // the zoneroot. This only currently works with "docker" VMs.
+            // the zoneroot. This does not work with HVMs.
```
papertigers, line 8700: nit: does not currently work with HVMs?

bahamat: I don’t want to set an expectation that someday it might. Even when/if p9fs comes about, I don’t know if this particular mechanism will be used with it. If it does someday/somehow change, I don’t think anybody will object because of this comment :-)

papertigers (approved):

PR #991, opened 2021-06-21
OS-8486 Ocassional issues when reprovisioning zones

blackwood821:
Issue 1:
```
Command failed: cannot mount 'zones/cores/<vm_uuid>': filesystem already mounted
```
This was happening frequently on a few of our CNs so I changed the code that unmounts the cores dataset to use the dataset name rather than the mountpoint and this seems to have resolved the issue. I have been unable to recreate the issue after the code change.

Issue 2:
```
Command failed: cannot mount 'zones/<vm_uuid>': mountpoint or dataset is busy
clone successfully created, but not mounted
```
This happens occasionally and I noticed there was already 1 retry in place from this commit: https://github.com/joyent/smartos-live/commit/b705c35be509cb4f0f65fff615ae226d8a02f3ab and most of the time that 1 retry works around the issue but occasionally additional retries are required. I modified the code to retry up to 10 times and it always seems to work within the first 3 retries.

Issue 3:
```
Error: Cannot start vm from state "provisioning", must be "stopped".
```
Occasionally `zone_state` is still `down` rather than `installed` while `state` is `provisioning`. I added up to 10 retries and it seems to always work by the first retry.

Issue 4:
```
Command failed: imgadm get: error (ImageNotInstalled): image "<image_uuid>" was not found on zpool "undefined"
```
Occasionally the VM GET response from `vminfod` does not include all fields (see https://gist.github.com/blackwood821/3cb90334fc2ead340a91d4fddf9ffb68 for more details), specifically the `zpool` field which is required to validate the image before reprovisioning. If the `zpool` field is unset I added a call to `VM.load()` and pass `"loadManually": true` which forces the call to bypass `vminfod` which seems to fix the issue.

I have tested thousands of reprovisions with these changes and they hold up in our environment.


blackwood821:
In my most recent test I ran 1,442 reprovisions in a row and 478 of those reprovisions encountered issue number 2 where the existing mount retry that was already in place worked around the issue but then there were 7 reprovisions where the first mount retry did not succeed and the second retry I added in did succeed. This shows that the extra retries are necessary though infrequent.

src/vm/node_modules/VM.js
```diff
@@ -10897,12 +10897,29 @@ exports.reprovision = function (uuid, payload, options, callback)
...
+                        log.warn('failed to load \"zpool\" field, retrying '
+                            + 'via manual load');
+
+                        VM.load(uuid, {
+                            log: log,
+                            loadManually: true
+                        }, function (err, obj) {
+                            if (err) {
+                                cb(err);
+                                return;
```
danmcd, line 10911: What do we do with "vmobj" in this case of failure? Feels like we'll leak something here.

src/vm/node_modules/VM.js
```diff
@@ -14683,12 +14714,57 @@ exports.start = function (uuid, extra, options, callback)
...
+                                    if (retryObj.zone_state === 'installed') {
+                                        vmobj = retryObj;
+                                        cb();
+                                        return;
+                                    }
+
+                                    if (retry_count < retry_max) {
+                                        _checkZoneState();
+                                    } else {
+                                        cb(err);
```
danmcd, line 14752: As earlier, what happens to retryObj in this case?

src/vm/node_modules/VM.js
```diff
@@ -10897,12 +10897,29 @@ exports.reprovision = function (uuid, payload, options, callback)
...
+
+                        VM.load(uuid, {
+                            log: log,
+                            loadManually: true
+                        }, function (err, obj) {
+                            if (err) {
+                                cb(err);
+                                return;
+                            }
+                            vmobj = obj;
```
danmcd, line 10913: Also, what do we do with the old version of vmobj?

danmcd (commented):
I have some nits and some potentially bigger issues.

It also feels like the four distinct issues you cite in this PR should be four distinct PRs, each an easily backed-out unit, in case something was missed.

danmcd:
Also, OS-8486 might be redundant given https://smartos.org/bugview/TRITON-1441

src/vm/node_modules/VM.js
```diff
@@ -11009,7 +11026,11 @@ exports.reprovision = function (uuid, payload, options, callback)
             });
         }, function (cb) {
             // unmount <zonepath>/cores so dataset is not busy
-            zfs(['umount', vmobj.zonepath + '/cores'], log,
+            // unmounting using the mountpoint fails often with:
+            //      "cannot unmount '/zones/<vm_uuid>/cores': not a mountpoint"
+            // so use the dataset name instead of the mountpoint which seems to
+            // solve the issue
+            zfs(['unmount', vmobj.zpool + '/cores/' + uuid], log,
```
bahamat, line 11033: The comment here is just not true. There have been literally hundreds of thousands of reprovisions where the alleged error does not occur. Stating that it happens "often" is at best, an overestimation.

The only reason the `not a mountpoint` error occurs is if the dataset is not mounted. And if the dataset is not mounted trying to unmount by the dataset name will return the error `not currently mounted`. So you've only exchanged one error message for another here.

Functionally, there is exactly zero difference here.

src/vm/node_modules/VM.js
```diff
@@ -14683,12 +14714,57 @@ exports.start = function (uuid, extra, options, callback)
                     err = new Error('Cannot start vm from state "'
                         + obj.state + '", must be "stopped".');
                     log.error(err);
-                    cb(err);
+
+                    if (obj.state === 'provisioning'
```
bahamat, line 14718: I'm really not clear on what this entire code block is supposed to be doing.

src/vm/node_modules/VM.js
```diff
@@ -11178,8 +11201,16 @@ exports.reprovision = function (uuid, payload, options, callback)
...
+                            '(retry + ' + (retry_count + 1) + ') mount '
+                                + 'results');
+
+                            retry_count++;
+
+                            if (err && retry_count < retry_max) {
+                                _retryBusy();
+                            } else {
+                                cb(err);
+                            }
```
bahamat, line 11213: Given that the symptom is "dataset busy", and the max retries is 10 (i.e., 10 seconds), maybe the best way to handle this is to simply retry 10s later instead of 1s rather than having to deal with retry counts.

Just changing `retry_delay` is a lot less invasive a change that I would feel more comfortable with.

bahamat (changes requested):

src/vm/node_modules/VM.js
```diff
@@ -10897,12 +10897,29 @@ exports.reprovision = function (uuid, payload, options, callback)
             vs.once('ready', function (ready_ev) {
                 try {
                     vmobj = vminfodEventGetVM(uuid, ready_ev);
+
+                    if (!vmobj.zpool) {
+                        log.warn('failed to load \"zpool\" field, retrying '
```
bahamat, line 10902: I think that OS-8316 fixed the underlying issue with vminfod here. This entire section can probably be reverted, at least until we can confirm that it's still ever an issue.

src/vm/node_modules/VM.js
```diff
@@ -11009,7 +11026,11 @@ exports.reprovision = function (uuid, payload, options, callback)
             });
         }, function (cb) {
             // unmount <zonepath>/cores so dataset is not busy
-            zfs(['umount', vmobj.zonepath + '/cores'], log,
+            // unmounting using the mountpoint occasionally fails with:
+            //      "cannot unmount '/zones/<vm_uuid>/cores': not a mountpoint"
+            // so use the dataset name instead of the mountpoint which seems to
+            // solve the issue
+            zfs(['unmount', vmobj.zpool + '/cores/' + uuid], log,
```
bahamat, line 11033: Another thing about this, is the "not a mountpoint" message is explicitly handled immediately below in the next stanza, but the error handling is not update.

This will turn a non-fatal error into a fatal error.

bahamat:
Superseded by #1078.

PR #1001, opened 2021-09-10
OS-8316 vminfod client shouldn't attempt to cleanup on a ready timeout

papertigers:


danmcd (dismissed):
Seems simple enough.

papertigers:
Turns out `_abort()` calls `stop()` which calls `_clearReadyTimeout()`.  
So I am opting for sticking on an additional property so that its easy to find with mdb.

papertigers:
cc @bahamas10 if you have time and wouldn't mind looking this over.

danmcd (approved):

bahamas10:
this change looks good to me! i like the idea.  one small nit: initialize `self.vs_ready_timeout_fired = false` when the object is created so it always exists.  otherwise, LGTM

danmcd:
Please update the bug report with testing done, and your recent analysis to change the approach mentioned above in this PR.

danmcd (approved):
Make sure testing & analysis goes in the bug report.  I can IA once it does.

PR #1006, opened 2021-11-10
OS-8334 Bring OpenSSL 3 to the platform

danmcd:
Only change needed for OpenSSL 3 is to increase the size of the ramdisk (again).


bahamat (dismissed):

danmcd:
I probably will wait to push this until the other two are ready, or else it's a zero-context push offering something that's not ready yet.  (The alternative is to file a distinct bug, but I think it's just better to wait.)

bahamat (approved):

PR #1007, opened 2021-11-11
TRITON-2264 COAL image needs to be "Other 64-bit" for compatibility with VMWare Fusion 12.2.x

bahamat:
Similar to joyent/sdc-headnode#57.


bahamat:
That’s a good question. It hasn’t because I don’t have an earlier version to test with.

But I have, in the past, run SmartOS VMs in Fusion using other-64 as the specified OS without issue. My understanding is that selecting a specific OS will enable some optimizations and that other-64 is expected to be the minimum baseline.

danmcd (approved):

PR #1008, opened 2021-11-16
OS-8337 illumos.sh generated by smartos-live must use native rpcgen

danmcd:
Tested on gcc10 and stock builds.


papertigers (approved):

danmcd:
Passes both master and gcc10 on local builds.  There's a jenkins job in https://jenkins.joyent.us/job/joyent-org/job/smartos-live/job/OS-8337/ that's holding this up, but this bug COULD be IA'ed if you think it's okay.

papertigers:
LGTM since gate now ships the tool

bahamat (approved):

PR #1013, opened 2022-02-13
OS-8353 piadm should get the download location from a config file

danmcd:
WORK IN PROGRESS.

### DIFF
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -176,11 +176,43 @@ vcurl() {
 	fi
 }
 
-# Well-known source of SmartOS Platform Images
+# Default well-known source of SmartOS Platform Images
 DEFAULT_URL_PREFIX=https://us-east.manta.joyent.com/Joyent_Dev/public/SmartOS/
 
-# Can be overridden by the user's PIADM_URL_PREFIX.
-URL_PREFIX=${PIADM_URL_PREFIX:-${DEFAULT_URL_PREFIX}}
+#
+# (Re)-Configure the default URL (and potentially other things in the future)
+# using 
+#
+config_check() {
+	# Can't do standalone_only per se as it errors out, but this only
+	# applies to standalone SmartOS.
+	if [[ "$TRITON_CN" == "yes" || "$TRITON_HN" == "yes" ]]; then
+		return
+	fi
+
+	if [[ -e /var/piadm/piadm.conf ]]; then
+		. /var/piadm/piadm.conf
+	else
+		# We're create /var/piadm/piadm.conf.
+		# NOTE: On an installation this will disappear as /var is
+		# on the ramdisk.
+		mkdir -p /var/piadm
+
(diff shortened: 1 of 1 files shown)


danmcd:
Testing notes in the ticket.

src/piadm.sh
```diff
@@ -176,11 +176,68 @@ vcurl() {
...
+		mkdir -p /var/piadm
+
+		# The latest default behavior should be hardcoded in this path.
+		# Update as we add more things.  Each change bumps
+		# PIADM_CONFIG_VERSION by one.
+		PIADM_CONFIG_VERSION=1
+		printf "PIADM_CONFIG_VERSION=%s\n" $PIADM_CONFIG_VERSION > \
+			$PIADM_CONF
+		printf "DEFAULT_URL_PREFIX=%s\n" $DEFAULT_URL_PREFIX >> \
+			$PIADM_CONF
```
bahamat, line 213: Consider a heredoc instead of multiple redirects.

src/piadm.sh
```diff
@@ -176,11 +176,68 @@ vcurl() {
...
+config_check() {
+	# Can't do standalone_only per se as it errors out, but this only
+	# applies to standalone SmartOS.
+	if [[ "$TRITON_CN" == "yes" || "$TRITON_HN" == "yes" ]]; then
+		return
+	fi
+
+	OLD_DEFAULT="$DEFAULT_URL_PREFIX"
+	if [[ -f $PIADM_CONF ]]; then
+		. $PIADM_CONF
```
bahamat, line 197: I prefer `source <file>` over `. <file>` because it's `grep`able.

bahamat (commented):
Does this still let you override with an environment variable?

danmcd:
> Does this still let you override with an environment variable?

Yes, see line 239.

src/piadm.sh
danmcd: I thought `source` was csh/tcsh/zsh only.

src/piadm.sh
danmcd: Like:
```
cat <<<FILE
PIADM_CONFIG_VERSION=$PIADM_CONFIG_VERSION
...
FILE > $PIADM_CONF
```
?

bahamat: Just `<<` not `<<<`, but otherwise yeah.

```
cat << FILE
stuff
FILE > $PIADM_CONF
```

src/piadm.sh
bahamat: It works in bash too.

danmcd: This file already uses `. <PATH>` elsewhere, not changing this one.

src/piadm.sh
danmcd: Will fix.

bahamat (approved):

PR #1016, opened 2022-03-10
OS-8363 Create 8gb image in smartos-live

danmcd:
Still under build, but should be VERY similar to OS-8250


bahamat (approved):

danmcd:
Testing notes in ticket.

PR #1022, opened 2022-04-07
Add workaround for 90 second LX systemd stop/reboot delay

smokris:
- When shutting down a SmartOS LX zone, `systemd-shutdown`'s `main()` calls `broadcast_signal()` first with SIGTERM, then with SIGKILL.
- At that point, the root cgroup becomes empty.
   - On recent real Linux kernels, systemd can use its "unified cgroup hierarchy" (cgroup-v2) codepath, [which signals when the cgroup becomes empty](https://github.com/systemd/systemd/blob/5645b4976ec35d24a341b6a7a3a7a205859e9864/src/core/cgroup.c#L3388), interrupting the `sigtimedwait()` and proceeding with the reboot without a delay.
   - On SmartOS LX, systemd uses its "legacy cgroup hierarchy" (cgroup-v1) codepath, [which _does not_ signal when the cgroup becomes empty](https://github.com/systemd/systemd/blob/5645b4976ec35d24a341b6a7a3a7a205859e9864/src/core/cgroup.c#L3412-L3413), so the `sigtimedwait()` continues until its 90 second timeout.

This PR modifies vmadm to send the `systemd-shutdown` process SIGCHLD periodically during a `vmadm stop` or `vmadm reboot` command, which causes `systemd-shutdown` to re-check its list of remaining processes, thereby removing the unnecessary delay.

Prior to this change, `vmadm stop` and `vmadm reboot` were taking slightly more than 90 seconds.  With this change, each takes about 10 seconds.

See discussion and @danmcd's initial testing on https://smartos.topicbox.com/groups/smartos-discuss/T7eaf4fad91e64ad8-M2e374d3cdfd9366191bc09b7.

### DIFF
--- a/src/vm/node_modules/VM.js
+++ b/src/vm/node_modules/VM.js
@@ -17641,11 +17641,16 @@ function doShutdownStop(vmobj, options, callback)
                 cb();
             });
         }, function waitInstalled(cb) {
-            VM.waitForZoneState(vmobj, 'installed', {timeout: STOP_TIMEOUT},
-                function shutdownWaitCb(err) {
+            var stop_timeout = Date.now(0) + STOP_TIMEOUT * 1000;
+            var ival = setInterval(function () {
+                // Every second, tell `systemd-shutdown` to re-check
+                // its list of remaining processes.
+                // (By default, `systemd-shutdown` waits 90 seconds
+                // for a root cgroup SIGCHLD that is never sent.)
+                killSig(vmobj.pid, 'SIGCHLD');
 
-                if (err && err.hasOwnProperty('code')
-                    && err.code === 'ETIMEOUT') {
+                if (Date.now(0) > stop_timeout) {
+                    clearInterval(ival);
 
                     log.info('Timeout waiting for shutdown to complete; '
                         + 'halting.');
@@ -17660,10 +17665,22 @@ function doShutdownStop(vmobj, options, callback)
                             }
                             cb(_err);
                         });
-                    return;
                 }
-                cb(err);
-            });
+
+                var load_opts = {fields: ['zone_state'], log: log};
+                vmload.getVmobj(vmobj.uuid, load_opts, function (_err, vmobj) {
+                    if (_err) {
+                        log.warn({err: _err, stdout: fds.stdout,
+                            stderr: fds.stderr}, 'Error checking status while halting zone');
+                        return;
+                    }
+
+                    if (vmobj.zone_state == 'installed') {
+                        clearInterval(ival);
+                        cb();
+                    }
+                });
+            }, 1000);
         }, function zonecfgNoAutoboot(cb) {
             zonecfg(vmobj.uuid, [unset_autoboot], {log: log},
                 function (err, fds) {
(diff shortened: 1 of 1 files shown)


danmcd (commented):
I haven't studied vmadm(1M) much, but how are you sure that vmobj.pid is the pid for systemd-shutdown?  What if you're sending SIGCHLD to something else?  What about older or alternate LX images that don't have modern systemd?

smokris:
Thanks for the feedback, @danmcd.

> how are you sure that vmobj.pid is the pid for systemd-shutdown?

If I understand correctly, `vmobj.pid` is [the PID of the `init` process for the zone](https://github.com/joyent/smartos-live/blob/master/src/vm/man/vmadm.8.md?plain=1#L1855).

During shutdown, `systemd` [replaces the `init` process image with `systemd-shutdown`](https://www.freedesktop.org/software/systemd/man/systemd-halt.service.html) (specifically, it [calls `execve()` without a `fork()`](https://github.com/systemd/systemd/blob/de7436b02badc82200dc127ff190b8155769b8e7/src/core/main.c#L1422)), reusing the same PID.

> What if you're sending SIGCHLD to something else? What about older or alternate LX images that don't have modern systemd?

Ah, good point.  To reduce that risk, I could have it check the process's `comm` and only send the signal if it's `systemd-shutdown`.

src/vm/node_modules/VM.js
```diff
@@ -17641,11 +17641,21 @@ function doShutdownStop(vmobj, options, callback)
...
         }, function waitInstalled(cb) {
-            VM.waitForZoneState(vmobj, 'installed', {timeout: STOP_TIMEOUT},
-                function shutdownWaitCb(err) {
+            var stop_timeout = Date.now(0) + STOP_TIMEOUT * 1000;
+            var ival = setInterval(function () {
+                // Every second, tell `systemd-shutdown` to re-check
+                // its list of remaining processes.
+                // (By default, `systemd-shutdown` waits 90 seconds
+                // for a root cgroup SIGCHLD that is never sent.)
+                fs.readlink('/proc/' + vmobj.pid + '/path/a.out',
```
danmcd, line 17650: Oh yes... I forgot to look at the bigger picture.  You know the pid already.

I had the man page section wrong, it's proc(5) now.  In illumos /proc/PID/argv, you get the information you need, albeit in a form that requires C-like parsing (strings with '\0' terminators).

danmcd: DAMN... I'm so sorry... this works exactly as it should for you.  This is an improvement.

danmcd (commented):
Sorry for my meandering into procfs.

You should test this on multiple kinds of LX Zone images.  I'm *happy* to build a test platform image for you that has this fix in it, OR you can use lofs tricks to mount yours over the one in use on the system.

Also, and maybe @bahamat might know, there may be a corresponding code segment in VMAPI that needs to have this as well.  (I can't remember if it's duplicate code or if VMAPI direclty uses files here.)

danmcd:
@bahamat and I had a chat offline... we're going to look a little deeper into this problem.  Your analysis of how systemd-shutdown behaves is EXCELLENT, but we're wondering if this fix belongs somewhere lower in the stack, so `zoneadm`(8) can DTRT as well.

danmcd:
I'm not sure if this requires fixes in both VM.js *and* in zone brand scripts.  This one may get delayed for a bit; I'm sorry.

bahamat:
I think if zoneadm handles this properly, vmadm should be fine as-is.

PR #1024, opened 2022-04-16
TOOLS-2541 Update slack channel

bahamat:
Org webhooks are also in place now so once this is created, jira should automatically get the PR link, and once it's merged Jenkins should auto build it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,6 +6,7 @@
 
 /*
  * Copyright 2022 Joyent, Inc.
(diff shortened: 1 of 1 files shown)


danmcd (approved):

danmcd:
Should you add MNX copyright or not yet?

danmcd:
IA is not here only because if you add MNX copyright I have to re-approve as well.

bahamat:
Oh, that’s right thanks.

danmcd (approved):

PR #1042, opened 2022-05-05
OS-8385 smartos-live should be smarter about setting MAX_JOBS and friends

danmcd:
This has been tested to prove that both illumos and other gmake get their jobs reduced to the output of the new optimize_jobs script.

### DIFF
--- a/configure
+++ b/configure
@@ -323,11 +323,15 @@ cat > "projects/illumos/illumos.sh" <<EOF
 NIGHTLY_OPTIONS="$nopts";			export NIGHTLY_OPTIONS
 GATE="${RELEASE_VER}";				export GATE
 CODEMGR_WS="$conf_root/projects/illumos";	export CODEMGR_WS
-MAX_JOBS=128
+MAX_JOBS=${MAX_JOBS:-128}
 maxjobs() {
         ncpu=\`kstat -p cpu_info:::state | grep -c on-line\`
 	zcpucap=\`pfexec mdata-get sdc:cpu_cap\`
-	zcpu=\`expr \$zcpucap / 100\`
+	if [[ \$zcpucap -eq 0 ]]; then
+		zcpu=\${MAX_JOBS}
+	else
+		zcpu=\`expr \$zcpucap / 100\`
+	vi
 	if [[ \$zcpu -lt \$ncpu ]]; then
 		expr \$zcpu + 2
         elif [[ \$(( \$ncpu + 2 )) -lt \${MAX_JOBS} ]]; then
--- a/configure
+++ b/configure
@@ -13,6 +13,7 @@
 
 #
 # Copyright 2021 Joyent, Inc.
+# Copyright 2022 MNX Cloud, Inc.
 #
 shopt -s xpg_echo
 unalias -a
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -322,10 +323,19 @@ cat > "projects/illumos/illumos.sh" <<EOF
...
 MAX_JOBS=128
 maxjobs() {
         ncpu=\`kstat -p cpu_info:::state | grep -c on-line\`
-        if [[ \$(( \$ncpu + 2 )) -lt \${MAX_JOBS} ]]; then
+	zcpucap=\`pfexec mdata-get sdc:cpu_cap\`
+	if [[ \$zcpucap -eq 0 ]]; then
+		zcpu=\${MAX_JOBS}
+	else
+		zcpu=\`expr \$zcpucap / 100\`
+	vi
```
bahamat, line 335: `vi` -> `fi` to close `if` statement

bahamat: Also, since the logic in this function is duplicated in tools/optimize_jobs, why not just call it instead?

configure
```diff
@@ -322,10 +323,19 @@ cat > "projects/illumos/illumos.sh" <<EOF
 NIGHTLY_OPTIONS="$nopts";			export NIGHTLY_OPTIONS
 GATE="${RELEASE_VER}";				export GATE
 CODEMGR_WS="$conf_root/projects/illumos";	export CODEMGR_WS
+# NOTE: build_illumos modifies this initial constant
 MAX_JOBS=128
 maxjobs() {
         ncpu=\`kstat -p cpu_info:::state | grep -c on-line\`
-        if [[ \$(( \$ncpu + 2 )) -lt \${MAX_JOBS} ]]; then
+	zcpucap=\`pfexec mdata-get sdc:cpu_cap\`
```
bahamat, line 330: While this seems to work as-is, if there is no cpu_cap value then zcpucap will be a zero length string, not the numeral `0`. There may be a subtle bug here waiting to lash out.

This also applies to tools/optimize_jobs below.

bahamat (changes requested):

configure
danmcd: "vi" --> damn good catch.

"Why not optimize_jobs?" --> Because `nightly.sh` is invoked by $illumos/usr/src/tools/nightly and it has no concept of smartos-live nor any of its subdirectories.  I'd prefer to keep it that way unless there's an existing compelling case of smartos-live awareness I'm missing.

configure
danmcd: Both will need fixing.

configure
bahamat: Oh, and this part is just creating illumos.sh. That also explains why everything seemed weirdly escaped. 👍

bahamat (approved):

tools/optimize_jobs
```diff
@@ -0,0 +1,41 @@
...
+else
+	zcpu=$(( $zcpucap / 100 ))
+fi
+
+if [[ $zcpu -lt $ncpu ]]; then
+	expr $zcpu + 2
+elif [[ $(( $ncpu + 2 )) -lt ${MAX_JOBS} ]]; then
+	expr $ncpu + 2
+else
+	printf "%d\n" \${MAX_JOBS}
```
nwilkens, line 38: Do we need the escape here?

bahamat: I think you’re right.

bahamat (approved):

nwilkens (approved):

PR #1043, opened 2022-05-06
OS-8382 Update README.md

jsadowyj0:
Updates README.md to show `git` rather than `scmgit`.


danmcd (commented):
Per Slack, you're going to have to be a bit more surgical about this change.  The rewhacking of '*' to '-' is going to mess up some set of MD rendering tools we may reconstitute here at MNX.

bahamat (approved):

PR #1045, opened 2022-05-19
OS-8386 vmware vmdk number of blocks is wrong

bahamat:


danmcd (dismissed):
Might not be an issue, but currently (cylinders * heads * sectors) == 3906000, so it's possible you may need to update these to reflect reality?

I will note that the transpostion of "3758" cylinders gets us much closer (and makes me wonder if it was a transposition error?!?).

Make sure there's testing noted.

bahamat:
It looks like what was originally done was assumed the img file was 2000000000.

I created a script to help out here. This is based on the [make_vmdk script in sdc-headnode](https://github.com/TritonDataCenter/sdc-headnode/blob/master/vmware/make_vmdk#L63-L69).

```
#!/bin/bash

g_image_bytes="${1:-516096}"

(( g_image_blocks = g_image_bytes / 512 ))
(( g_image_cylinders = g_image_blocks / (16 * 63) ))

printf 'Size is  : %d\n' "$g_image_bytes"
printf 'Blocks   : %d\n' "$g_image_blocks"
printf 'Cylinders: %d\n' "$g_image_cylinders"
```

Starting with 2000000000, this gives us the values that are currently in the vmdk file:

```
> ./sizecalc 2000000000
Size is  : 2000000000
Blocks   : 3906250
Cylinders: 3875
```

Using the actual size of the image, we get this instead:

```
> ./sizecalc 1940000256
Size is  : 1940000256
Blocks   : 3789063
Cylinders: 3758
```

So it looks like `3875` isn't a typo, it just started with a different value. But it does also mean that I should fix that value in the PR.

danmcd (approved):
Brilliant! Good analysis and software archeology too.

PR #1049, opened 2022-07-11
OS-8394 Move smartos-live (and friends if needed) to pkgsrc.smartos.org

danmcd:


src/smartdc/bin/pkgsrc-setup.sh
```diff
@@ -20,16 +21,17 @@ fi
 root="/${1}"
 
 # Occasionally, the filename and hash will need to be updated. Refer to
-# pkgsrc.joyent.com for changes. Old boostrap tars are kept around indefinitely,
+# pkgsrc.smartos.org for changes. Old boostrap tars are kept around indefinitely,
 # so there's no particular urgency for getting it done when a new bootstrap
 # is available, but we'll want to stay relatively up to date.
+# XXX KEBE ASKS -- change these?!?
```
danmcd, line 27: Maybe not?  But I'd like an authoritative answer (and maybe a commitment to make sure `bootstrap-trunk-tools...` is on pkgsrc.smartos.org?).

jperkin: The trunk bootstrap tarballs have a date in them, and yes are available on pkgsrc.smartos.org.

The LTS tarballs do not have a date so are very occasionally updated in-place, like I've done recently for the move to pkgsrc.smartos.org, but are otherwise left untouched for the duration of the release.

So yes, using pkgsrc.smartos.org is correct now, and depending on whether you're using 2021Q4 or trunk just grab whatever is available from https://pkgsrc.smartos.org/packages/SmartOS/bootstrap/ and start using it as they've all been updated to be correct.

danmcd: Cool.  I'll lose this KEBE comment, and promote this.  I'll note that illumos-joyent's usr/src/test has a similar set of variables:

```
    # We should always use the same pkgsrc version as we have installed
    # on the build machine in case any of our tests link against libraries
    # in /opt/local
    PKGSRC_STEM="https://pkgsrc.smartos.org/packages/SmartOS/bootstrap"
    BOOTSTRAP_TAR="bootstrap-2018Q4-tools.tar.gz"
    BOOTSTRAP_SHA="b599667c80e4a42157763ed25d868ec7dc34962d"
```

that I'll put in an illumos-joyent PR.

jperkin (dismissed):

danmcd:
Testing notes in ticket.

bahamat (approved):

PR #1053, opened 2022-07-27
OS-8400 pkgsrc-setup should be able to re-bootstrap

bahamat:


danmcd:
1.) There's no man page or usage information you need to update with this new `--clean` flag? (Blocking +1 for review.)
2.) Test notes in the ticket?  (Blocking IA)

src/smartdc/bin/pkgsrc-setup.sh
```diff
@@ -18,6 +18,21 @@ if [[ -n "$TRACE" ]]; then
     set -o xtrace
 fi
 
+# If there's an existing install...
+if [[ -e /opt/tools/bin/pkgin ]]; then
+    # Re-bootstrap pkgsrc if requested.
+    if [[ $1 == '--clean' ]]; then
+        pkg_stash=$(mktemp)
+        pkgin export > "$pkg_stash"
+        rm -rf /opt/tools
```
danmcd, line 27: Remove the $pkg_stash file too?

travispaul (approved):

src/smartdc/bin/pkgsrc-setup.sh
danmcd: Ooops, my bad. Taken care of later.

danmcd:
@travispaul put some test notes in the ticket.

danmcd:
And I answered my own question about man pages or other docs.

danmcd (approved):

man/smartdc/man/man1/pkgsrc-setup.1.md
```diff
@@ -0,0 +1,25 @@
+# pkgsrc-setup(1) -- bootstrap pkgsrc-tools
+
+
+## SYNOPSIS
+
+`pkgsrc-setup [--clean]`
+
```
danmcd, line 7: There's no mention of /path/to/root as an option.  I take it that's deliberate and not an oversight?

bahamat: Added a man page

bahamat: Correct. The path to root is only intended to be used during smartos install when the zpool is mounted to an alternate location. If you just tried to install to an arbitrary location I don't even know if the installation would be usable. On a running system, any location other than `/opt/tools` is not supported.

danmcd (approved):
Choosing to keep the /path/to/root undocumented is okay.

travispaul (approved):

PR #1054, opened 2022-07-28
OS-8400 pkgsrc-setup should be able to re-bootstrap (fix Makefile)

bahamat:


travispaul (approved):

danmcd (approved):
Sorry I didn't catch this.  it built locally from a fresh pull + this fix.

PR #1055, opened 2022-07-30
Increase maxBuffer for imgadm zfs listing

drscream:
The maxBuffer is not suitable for production environments with lot's of snapshots and delegate datasets. Especially when the structure contains lot's of sub-folders.

### DIFF
--- a/src/img/lib/imgadm.js
+++ b/src/img/lib/imgadm.js
@@ -856,7 +856,7 @@ IMGADM.prototype._loadImages = function _loadImages(callback) {
             log: self.log,
             errMsg: 'could not load images',
             execOpts: {
-                maxBuffer: 10485760  /* >200k hit in prod, 10M should suffice */
+                maxBuffer: 20971520  /* >200k hit in prod, 20M should suffice */
             }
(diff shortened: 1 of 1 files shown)


bahamat:
Can you provide some metrics on behavior before and after?

drscream:
Example `zfs list` output with `wc -c` on one of our production systems:

```
$ /usr/sbin/zfs list -t filesystem,volume,snapshot -pH -o name,origin,mountpoint,imgadm:ignore | wc -c
10808528
```

As `10808528` is larger than `maxBuffer` value `10485760` the `imgadm list` command fail with a strange output of listed snapshots, datasets. After I increased the value the output is working again.

The performance of `imgadm` is already really bad on that system:

```
$ time imgadm list
[...]

real    5m51.537s
user    0m5.972s
sys     1m55.309s

```

bahamat:
How many datasets do you have?

drscream:
```
$ zfs list | wc -l
1200
```

The server receive datasets into zones via `znapzend` as a backup solution.

danmcd (approved):

PR #1056, opened 2022-09-01
OS-8405 Have Jenkinsfile use parallel stages

danmcd:
WIP, thinking that parallel stage builds (which can be done as they're all self-contained) might be a new parameter for build-with-parameters.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -115,6 +115,7 @@ pipeline {
 		'(false/DEFAULT).</p>'
         )
     }
+    def stages_to_run = [:]
     stages_to_run['default'] = {
             stage('default') {
                 agent {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -107,233 +107,261 @@ pipeline {
...
                 sh('''
 set -o errexit
 set -o pipefail
-export MANTA_TOOLS_PATH=/root/bin/
-./tools/build_jenkins -c -F strap-cache -S strap-cache
+./tools/build_jenkins -c -F check
                 ''')
             }
             post {
+                // We don't mattermost-notify here, as that doesn't add much
```
bahamat, line 333: Remove mattermost reference. Maybe "we don't chat notify" or just "we don't notify" (maybe we'll use something other than chat someday)

danmcd: Ooops, was too specific...

bahamat (approved):

PR #1061, opened 2022-11-15
OS-8422 Give gcc10 (alternate-compiler) its own directory in .../public/builds/.

danmcd:
Not intended for 20221117, but time to take a look.

### DIFF
--- a/tools/build_jenkins
+++ b/tools/build_jenkins
@@ -126,7 +126,7 @@ while getopts "cdhF:S:" opt; do
         d)
             # build debug platform bits
             ENGBLD_CONFIGURE_DEBUG_ARG='-d'
-            ENGBLD_DEBUG_SUFFIX='-debug'
+            ENGBLD_DEBUG_SUFFIX=${PLATFORM_DEBUG_SUFFIX:--debug}
             ;;
         F)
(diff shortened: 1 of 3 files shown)


danmcd:
See line 161-ish of build_jenkins --> we can eliminate the sleep for the non-default/non-debug build unless one thinks having a different-enough stamp would be worth it.

tools/build_jenkins
```diff
@@ -28,6 +28,9 @@ if [[ -z "${ENGBLD_DEST_OUT_PATH}" ]]; then
     export ENGBLD_DEST_OUT_PATH=/public/builds
 fi
 
+export TIMESTAMP=$(TZ=UTC /bin/date "+%Y%m%dT%H%M%SZ")
+export BUILDSTAMP=${TIMESTAMP}
+
```
bahamat, line 33: Does this still need to be done early?

bahamat (approved):

tools/build_jenkins
danmcd: Unsure if it needs to be done early or not, but it's not strictly necessary. For this change I also reverted all of OS-8408 (60-sec sleep for gcc10) so prior to that the BUILDSTAMP was set this early (and for all of our Jenkins time prior to 8408).

PR #1062, opened 2022-11-18
TOOLS-2574 Purge dependency on us-east.manta.joyent.com name

danmcd:
This may be an existing bug, hence the vague branch name and OS-xxxx.  This is slipshod right now, and I may either be overreaching OR missing things.


src/vm/lib/metadata/agent.js
```diff
@@ -21,6 +21,7 @@
  * CDDL HEADER END
  *
  * Copyright 2019 Joyent, Inc.
+ * Copyright 2022 MNX Cloud, Inc.
```
danmcd, line 24: Will need copyright update.

src/vm/node_modules/macaddr/package.json
```diff
@@ -26,8 +26,8 @@
     "test": "istanbul cover --print none test/runall.js"
   },
   "author": {
-    "name": "Joyent",
-    "url": "joyent.com"
+    "name": "Triton Data Center",
```
danmcd, line 29: Worth a version bump?

src/vm/node_modules/triton-netconfig/node_modules/ip6addr/package.json
```diff
@@ -19,8 +19,8 @@
     "test": "istanbul cover --print none test/runall.js"
   },
   "author": {
-    "name": "Joyent",
-    "url": "joyent.com"
+    "name": "Triton Data Center",
```
danmcd, line 22: Version bump?

src/vm/node_modules/triton-netconfig/package.json
```diff
@@ -4,21 +4,21 @@
   "description": "Common methods for managing Triton network configuration",
```
danmcd, line 4: Version bump worthy?

src/vm/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vmadm",
   "description": "administrative tool(s) for managing VMs",
   "version": "1.0.5",
```
danmcd, line 4: Version bump?

danmcd:
This has built successfully in the PR-1062 Jenkins job.  Beyond that it'll need some testing.

src/img/node_modules/imgmanifest/package.json
```diff
@@ -3,8 +3,8 @@
   "description": "Library for working with Triton/SDC/SmartOS image manifests",
   "version": "3.1.0",
```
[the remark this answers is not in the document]
travispaul: This has already been updated in the source repo: https://github.com/TritonDataCenter/node-imgmanifest

I'm not sure if we have a special process for updating the modules in this repo since we commit the node_modules directory in this repo.

The correct answer for this package.json will be true for every package.json within node_modules, I suspect most, if not all, have been updated in their source repo.

src/fw/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "fw",
   "description": "Administrative tool for managing SmartOS VM firewalls",
   "version": "1.2.2",
```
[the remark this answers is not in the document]
travispaul: It looks like we don't publish this to NPM or use it elsewhere AFAICT so I don't think it matters much, probably wouldn't hurt to bump patch though while where here.

src/vm/package.json
travispaul: This looks like the same situation as the `fw` module. I don't believe it's used outside of smartos live, a version bump wouldn't hurt but I don't think it's necessary either.

danmcd:
Successfully built locally with default, debug, and gcc7.  Jenkins PR-1062#11 will confirm it on the Jenkins side.

PR #1063, opened 2022-11-18
OS-xxxx use more content-type: in manta uploads

danmcd:


tools/build_ctftools
```diff
@@ -133,7 +162,9 @@ function upload
 	verbose mmkdir -p "$mbdir"
 
 	for file in $bits_dir/*; do
-		verbose mput -f "$file" "$mbdir/$(basename $file)"
+		filename="$(basename $file)"
+		verbose mput -H "$(content_type $filename)" -f "$file" \
```
bahamat, line 166: This needs to be:

    -H "content-type: $(content_type $filename)"

bahamat (changes requested):

tools/build_ctftools
bahamat: Oh, wait...you're already including the header name. That *should* work. I guess we'll need to dig into Jenkins to see what went wrong.

danmcd: I think it's TritonDataCenter/eng that needs some changes here too.  Building one right now that I hope will include the content-type branch of eng.

danmcd: In the "upload-bits" script.

danmcd: See TritonDataCenter/eng#62

PR #1064, opened 2022-11-21
TRITON-2304 New image server names

danmcd:
I think we forgot to update the submodule(s) here.


bahamat (approved):
If this is

danmcd:
If the PR-1064 build finishes w/o issue(s) and the artifacts are there, I'll push this.

PR #1065, opened 2022-11-30
OS-8428 sysinfo may fail on strings "sanitized" by smbios but are toxic to json

bahamat:


danmcd (dismissed):
We'll be launching node 8 times per invocation after this change.  I'm okay with it for now because correctness overrides performance, but please make a note of this (perhaps in lines 122-124?) in case it bites us in the ass later.

I don't know how often sysinfo is invoked and how long it's expected to run.

danmcd (approved):
Nice!!!

bahamat:
@danmcd Ok, after your comment I did some benchmarking and it turns out that calling node 8 times increased the run time by 0.5s-1s.

I decided to try rewriting it. I ran the original version and my new version each 100x. The old version runs on average `1.210`s, with stdev of `0.013`s. The new version now runs on average `1.258`s, with stdev of `0.011`.

So of my tests with this new version, we're on average 1/20ths slower.

bahamat:
Customer reports that this version also works.

danmcd:
Adding IA based on email.  See if someone else will review if possible.

travispaul (approved):

PR #1068, opened 2022-12-31
OS-8428 sysinfo may fail on strings "sanitized" by smbios but are toxic to json (fix CPU_Type)

bahamat:


pfmooney:
I encountered this on a machine running recently-updated qemu/kvm/ovmf bits, which seems to emit bogus SMBIOS data for the guest:
```
1024  67   SMB_TYPE_PROCESSOR (type 4) (processor)

  Manufacturer: \x01
  Version: \x01
  Asset Tag: AMD
  Location Tag: \x01
```

The data for Manufacturer would land in the `sysinfo` field for `CPU Type`, causing `vminfod` and friends to be upset over the invalid json. 

With this fix in place (caching now-valid data), `vminfod` would run again.

src/sysinfo
```diff
@@ -196,6 +196,12 @@ function get_smartos_cpu_info()
         if [[ -z ${CPU_Version} ]]; then
             CPU_Version="Unknown"
         fi
+        # Platforms that include illumos#13560, smbios  may emit "sanitized"
+        # strings that are not valid json. So we need to pass these through
+        # JSON.stringify.
+        eval $(/usr/node/bin/node -e '
+            console.log("CPU_Version=" + JSON.stringify("'"$CPU_Version"'"))
+        ')
```
bahamat, line 204: This is the same fix as in 27dee23aaa6837651d8c1d1a376f308344100aca, it's just a different call to smbios.

There are no other calls to smbios where the value is used in the sysinfo output.

danmcd (approved):

danmcd:
Also adding IA because you have noted your tests in the ticket.

PR #1076, opened 2023-08-17
OS-8469 Update default sshd_config to reduce log spam

bahamat:
This only affects the global zone because base zone images have their own sshd_config.


danmcd (commented):
Do we wish to update similar diffs in illumos-extra as well? Or is Just Fixing it Here sufficient   I can do the illumos-extra portions if need be.

Meanwhile to REDUCE the diffs between smartos-live and the existing illumos-extra, please also:

- Comment-out `GatewayPorts` line (104); it defaults to no.
- Delete the commented-out `#UseLogin` line which is right above the UsePrivilegeSeparation you're deleting already.

danmcd (approved):

travispaul (approved):

PR #1077, opened 2023-09-01
OS-8484 Headnode NTP config should be able to support RAN networks

bahamat:
Companion: TritonDataCenter/illumos-joyent#453

### DIFF
--- a/src/smartdc/lib/ntp_config.js
+++ b/src/smartdc/lib/ntp_config.js
@@ -2,6 +2,27 @@
 /* vim: syn=javascript ts=8 sts=8 sw=8 noet: */
 
 /*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License, Version 1.0 only
+ * (the "License").  You may not use this file except in compliance
+ * with the License.
+ *
+ * You can obtain a copy of the license at http://smartos.org/CDDL
+ *
+ * See the License for the specific language governing permissions
+ * and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file.
+ *
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ *
  * Copyright 2019 Joyent, Inc.
  * Copyright 2023 MNX Cloud, Inc.
  */
@@ -206,7 +227,7 @@ generate_file(servers, allowed_subnets, trusted, orphan)
 			}
 
 			/*
-			 * We should deduplicate these networks since the -a
+			 * Now deduplicate these networks since the -a
 			 * flag can be passed multiple times.
 			 */
 			if (seen.indexOf(as[0] + '/' + mask) === -1) {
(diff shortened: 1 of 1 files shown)


src/smartdc/lib/ntp_config.js
```diff
@@ -199,9 +205,16 @@ generate_file(servers, allowed_subnets, trusted, orphan)
 				process.exit(1);
 			}
 
-			out.push(mod_util.format('%s %s mask %s',
-			    fam === 4 ? 'restrict' : 'restrict -6', as[0],
-			    mask));
+			/*
+			 * We should deduplicate these networks since the -a
```
danmcd, line 209: Is this a "We don't do it now, but we should," sort of comment?  If so, please mark with TODO:

```
/*
 * TODO: We should deduplicate...
```

If this code below actually does the dedup (doesn't *seem* to but I might be missing something), lose the "should" as we're actually deduplicating:

```
/*
 * We deduplicate these networks...
```

src/smartdc/lib/ntp_config.js
```diff
@@ -1,6 +1,11 @@
 #!/usr/node/bin/node
 /* vim: syn=javascript ts=8 sts=8 sw=8 noet: */
 
+/*
```
danmcd, line 4: Also, is this under some sort of license (CDDL1.0 or MPL2.0)?

src/smartdc/lib/ntp_config.js
bahamat: Clarified this comment.

src/smartdc/lib/ntp_config.js
bahamat: Added CDDL license header.

bahamat: Also, CDDL vs MPL here:

The reason for CDDL, with smartos-live, the convention has been for files that natively live in the repo (i.e., are not copied in from elsewhere), things tend to be CDDL. This file predates Joyent's significant use of MPL, so it seems the original intention would have been CDDL.

src/smartdc/lib/ntp_config.js
```diff
@@ -1,6 +1,32 @@
...
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file.
+ *
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ *
+ * Copyright 2019 Joyent, Inc.
```
danmcd, line 26: Last nit -> Copyright block is its own block after license, typically. Also the CDDL 1.0 text is shorter, see $ILLUMOS/$SRC/prototypes/prototype.c for the example.

danmcd (commented):
One last nit.  WOuld approve otherwise, thanks for the clarification.

src/smartdc/lib/ntp_config.js
bahamat: Oh, ok. I just `cat`'d a file at random and did it the same way.

danmcd: Upon further review... other .js files follow this convention (from way back I'd guess).

danmcd (approved):

src/smartdc/lib/ntp_config.js
bahamat: After looking at the rest of the repo, *most* files use the "expanded" CDDL header that I used here, and there are only a handful with separate copyright block. After discussing on Slack, we've decided that it's ok to leave this as-is.

PR #1078, opened 2023-09-19
OS-8486 Occasional issues when reprovisioning zones

bahamat:
This should help with some of the most commonly reported reprovision failures. Overall, reprovisioning is extremely stable. There are occasional failures experienced by almost everyone, and there is one reported case where failures are frequent. This should address the vast majority of those cases. This may not (and likely doesn't) address *all* reprovisioning failures, and failures being extremely rare it makes this difficult to accurately represent the degree of change.

But overall, this will hopefully reduce the incidents of reprovision failure experienced by operators.

See [OS-8486][1] for testing notes.

[1]: https://smartos.org/bugview/OS-8486

[OS-8486]: https://mnx.atlassian.net/browse/OS-8486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/vm/node_modules/VM.js
+++ b/src/vm/node_modules/VM.js
@@ -10987,8 +10987,8 @@ exports.reprovision = function (uuid, payload, options, callback)
                     VM.waitForZoneState(l_obj, 'installed',
                         {timeout: STOP_TIMEOUT},
                         function reprovisionStopWaitCb(err) {
-                            if (err && err.hasOwnProperty('code') &&
-                                err.code === 'ETIMEOUT') {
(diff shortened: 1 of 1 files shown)


danmcd (approved):
Having seen this in private before, I see nothing changed to alter my then-assessment...

travispaul (approved):

danmcd:
Thank you for documenting the tests both done and not-done.

PR #1079, opened 2023-10-11
LINUXCN-26 lsblk output has changed in Debian 12

travispaul:
The `mountpoint` property on block devices is now `mountpoints` and is also an array.

If no mountpoints exist, the value is: `[null]`

Also filter out any zvol devices which show up in `lsblk` after a Linux CN has been setup.

See [LINUXCN-26](https://mnx.atlassian.net/browse/LINUXCN-26) for testing notes.

[LINUXCN-26]: https://mnx.atlassian.net/browse/LINUXCN-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/disklayout.js
+++ b/src/disklayout.js
@@ -67,7 +67,7 @@ dolayout(disks, layout, nspares, excluded, enable_cache, width)
 			});
 
 			return (mountpoints.length === 0);
-		})
+		});
 	}
 
 	config = disklayout.compute(disks, layout, nspares, enable_cache,
--- a/src/disklayout.js
+++ b/src/disklayout.js
@@ -47,7 +47,7 @@ dolayout(disks, layout, nspares, excluded, enable_cache, width)
 			return (true);
 		});
 	} else if (os.type() === 'Linux') {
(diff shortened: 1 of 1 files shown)


src/disklayout.js
```diff
@@ -46,11 +47,26 @@ dolayout(disks, layout, nspares, excluded, enable_cache, width)
 			return (true);
 		});
 	} else if (os.type() === 'Linux') {
+		var zvolExp = /^zd[0-9a-z]*$/;
```
danmcd, line 50: If the device is "zd" all by itself, it'll match this expression.  Is that desirable? If not, change the `*` to `+`.

travispaul: Good point, I've pushed that change, thanks!

danmcd (approved):
After internal discussion I'm good.

bahamat (approved):

PR #1081, opened 2023-11-13
OS-8496 Deprecate docker registry access

bahamat:


src/img/lib/cli.js
```diff
@@ -783,11 +783,6 @@ CLI.prototype.do_sources.options = [
         helpArg: '<source>',
         help: 'Add a source. It is appended to the list of sources.'
     },
-    {
-        names: ['add-docker-hub'],
-        type: 'bool',
-        help: 'A shortcut for "imgadm sources -t docker -a https://docker.io".'
-    },
```
danmcd, line 790: This doesn't have an sdc-imgapi counterpart?

danmcd:
Also please add testing notes to the ticket.

src/img/lib/cli.js
bahamat: Nope. This is purely for importing images into SmartOS.

travispaul:
The `add-docker-hub` command is mentioned in a few other places including some tests in smartos-live. Do any of those references need to be updated/removed? 

https://hound.tritondatacenter.com/?q=add-docker-hub&i=nope&files=&repos=

danmcd (approved):
Need testing notes for IA.

danmcd (approved):

danmcd:
Please make sure the node-imgadm gets its issues addressed as well.

bahamat:
> The add-docker-hub command is mentioned in a few other places including some tests in smartos-live. Do any of those references need to be updated/removed?

The only place it's pertinent is in img/lib/cli.js and the man page. I'm only removing the documentation for it so that people will stop using it, but I'm not removing the feature. You can still use the `--add-docker-hub` flag, and add sources with type `docker`. In the future I might remove those as well, but we'll always need to handle the existence of docker sources, if only to ignore them.

danmcd (commented):
Re-upping.

danmcd (approved):
Re-upping.

PR #1085, opened 2023-12-06
OS-8506 Get ready for illumos#16099 changes to err.h

bahamat:


danmcd:
Adding IA as I proved the ostensibly harder of the two (strap-cache build on updated system) worked.
Since this is just a smartos-live change, push at your discretion.

PR #1088, opened 2024-01-22
OS-6789 vmadm recv does not create cores dataset

bahamat:


src/vm/node_modules/VM.js
```diff
@@ -8799,10 +8799,23 @@ function installZone(payload, log, callback)
...
-            // when receiving zoneinit is never run.
+            // When receiving zoneinit is never run, instead detach/attach the
+            // zone to ensure the cores dataset gets created properly by the
+            // brand scripts.
             if (receiving) {
-                cb();
-                return;
+                vasync.pipeline(
+                    {funcs: [
+                        function (_, cb2) {
```
danmcd, line 8808: These are in a vasync.pipeline so they will run sequentially, right?

src/vm/node_modules/VM.js
```diff
@@ -8799,10 +8799,23 @@ function installZone(payload, log, callback)
...
+                            zoneadm(['-u', payload.uuid, 'detach'], log, cb2);
+                        },
+                        function (_, cb2) {
+                            zoneadm(['-u', payload.uuid, 'attach'], log, cb2);
+                        }
+                    ]}, function (e) {
+                        cb(e);
```
danmcd, line 8815: I assume passing (e) into cb() (vs. the old way of calling cb() with no args) will not break anything?

danmcd (commented):
*this close* to approval.  I just need answers to things that are likely due to my relative ignorance of JS.

src/vm/node_modules/VM.js
bahamat: Correct.

src/vm/node_modules/VM.js
bahamat: That’s a good question. I don’t *know* that. But I couldn’t get into a situation where it would fail.

In any case, I think that if something *does* go wrong it’s better to try to float that up than it would be to silently ignore it.

The only thing I can think of that would cause it to fail is like out of memory and can’t fork (which, the vm would later fail to boot anyway), or an out of space error trying to create the dataset (again, the vim would be unable to boot later). I can’t think of any *recoverable* reasons this might fail.

bahamat: Also, passing `(e)` into `cb()` is just sending the error (if any) into the callback of the function that called us. The previous behavior was doing nothing, so there couldn't be an error to pass.

With JS in general (especially in ES5 like we're using here), it's pretty much standard to pass your errors into the callback function. Eventually there should be a handler that does something about it. We do this another 48 times at various points throughout VM.js, so that's definitely the right thing to do because we're pretty deep in the call stack when we get to this point.

danmcd (approved):
Thank you for answering my questions.

danmcd:
I had such problems sending & receiving bhyve outside the scope of these changes that I'm opting to not worry about it.

PR #1090, opened 2024-02-15
OS-8523 vmadm should worry less about bootrom

bahamat:


danmcd (commented):
Seem sensible at first glance.

danmcd:
Will grant IA once @travispaul signs off (pardon the cart-before-horse earlier...).

travispaul (approved):

PR #1096, opened 2024-03-30
OS-8536 piadm default config should not include a hostname override

bahamat:


src/piadm.sh
```diff
@@ -206,6 +206,12 @@ config_check() {
 		return
 	fi
 
+        if grep -q us-east.manta.joyent.com $PIADM_CONF ; then
+            # If they have a broken config with the stale joyent name,
+            # nuke it.
+            rm "${PIADM_CONF:?}"
+        fi
```
bahamat, line 213: I'm not entirely sure we should proactively delete it. SmartOS folks will need to delete it themselves before it will work again. Triton users may not ever know the difference.

I'm still soft inclined to keep it for Triton users.

The down side of this is that if someone wants to use us-east.manta.joyent.com, we're explicitly not letting them. Also, maybe this should be more targeted?

danmcd: IMHO you should be verbose about it, even without -v.  ANd nuke it.

danmcd: Still waiting on your opinion of verbosity.  Please defend lack thereof, or make this verbose.

bahamat: Updated to print a warning to stderr.

danmcd (dismissed):

danmcd:
Pleaste put test notes in ticket for an IA.

danmcd:
Actually... I did the testing myself!

src/piadm.sh
```diff
@@ -206,6 +206,14 @@ config_check() {
 		return
 	fi
 
+        if grep -q us-east.manta.joyent.com $PIADM_CONF ; then
+            # If they have a broken config with the stale joyent name,
+            # nuke it.
+            printf 'WARNING: Removing stale image server ' >&2
+            printf '(us-east.manta.joyent.com)' >&2
```
danmcd, line 213: Needs a newline at the end.  Otherwise this works!

danmcd (approved):

PR #1100, opened 2024-04-08
OS-8540 piadm should not emit unknown_fstyp messages

danmcd:


travispaul (approved):

src/piadm.sh
```diff
@@ -627,13 +627,14 @@ update_boot_sectors() {
...
 		# System Partition (ESP)
-		type=$(fstyp "/dev/dsk/${boot_devices[0]}s0")
+		type=$(fstyp "/dev/dsk/${boot_devices[0]}s0" 2>/dev/null)
 		if [[ "$type" == "pcfs" ]]; then
 			# If we detect PCFS on s0, it's LIKELY an EFI System
 			# Partition that was crafted manually.  Use s1 if it's
 			# ZFS, or bail if it's not.
 
-			s1type=$(fstyp "/dev/dsk/${boot_devices[0]}s1")
+			s1type=$( \
```
bahamat, line 636: The backslash is unnecessary here, and I think the enclosing parens should be styled like we would for enclosing braces.

E.g.:

```
> cat t.sh 
#!/bin/bash

d=$(
  date
)

echo "$d"
> ./t.sh 
Mon Apr  8 10:41:18 PDT 2024
```

bahamat (changes requested):

src/piadm.sh
danmcd: I think I fixed it.

bahamat (approved):

travispaul (approved):

Issue #1101, opened 2024-05-07
Build Error


discostur:
Hi,

i just tried to build SmartOS from source. Everything is setup and im building inside a base-64-lts 21.4.1 zone running on SmartOS 5.11 joyent_20220313T010148Z. During the gmake world i get: 

```
smartosbuild.intern.internetx.de --> Job output
/root/smartosbuild/smartos-live/projects/illumos/usr/src/tools/sgs
/root/smartosbuild/smartos-live/projects/illumos/usr/src/tools/sgs/include
/root/smartosbuild/smartos-live/projects/illumos/usr/src/tools/sgs/sgsmsg
[...]

PR #1102, opened 2024-05-08
smartos-live#1101 Minimum platform image to build SmartOS is 20220807

teutat3s:
See #1101

- bump the `base-64-lts` image version to `21.4.1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,33 +165,32 @@ QEMU
 ## Setting up a Build Environment
 
 The first step when building is to set up a build environment. The SmartOS
-build requires building on SmartOS.  As of the `base-64-lts 21.4.0` build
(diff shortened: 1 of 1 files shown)


danmcd (approved):
Unsure if we need to open an OS ticket for this or not.  A pleasant side-effect of this also is that we address TOOLS-2574, so maybe we fix this as part of that bug instead?

bahamat:
> Unsure if we need to open an OS ticket for this or not. A pleasant side-effect of this also is that we address TOOLS-2574, so maybe we fix this as part of that bug instead?

@danmcd I've set title of this PR to reference #1101, which is sufficient for ticketing reference.

bahamat (approved):

PR #1104, opened 2024-06-17
OS-8552 illumos-joyent should use tools yacc per illumos#16319

danmcd:
NOTE:  This commit cannot land until https://www.illumos.org/issues/16319 lands and is merged into illumos-joyent.  It's a baby build flag day for us.

### DIFF
--- a/configure
+++ b/configure
@@ -13,7 +13,7 @@
 
 #
 # Copyright 2021 Joyent, Inc.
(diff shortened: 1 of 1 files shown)


danmcd:
This also means automated builds will fail.

jperkin (approved):

PR #1105, opened 2024-06-26
OS-8518 Receiving a BHYVE VM fails due to premature ZFS dataset creation

danmcd:
Seems like a big hammer, and the answer might be to not prematurely create a ZFS dataset.

### DIFF
--- a/src/vm/node_modules/VM.js
+++ b/src/vm/node_modules/VM.js
@@ -11650,7 +11650,10 @@ function getAllDatasets(vmobj)
         disk = vmobj.disks[disk];
         if (disk.hasOwnProperty('zfs_filesystem')) {
             disk_dataset = disk.zfs_filesystem;
-            // Don't push if child of top-level zfs_filesystem.
+            // Child datasets of the zone root are automatically
+            // included because of -r (recursive).  If the dataset
+            // name is not a child of the zone root, then add it to
+            // the list of datasets we need to transfer.
             if (topmost_dataset !=
                 disk_dataset.substr(0, topmost_dataset.length)) {
                 datasets.push(disk.zfs_filesystem);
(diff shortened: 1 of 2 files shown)


danmcd:
Test notes in ticket (and ticket is public too on https://smartos.org/bugview/OS-8518 ).

tsoome (dismissed):

src/vm/node_modules/VM.js
```diff
@@ -11633,15 +11634,27 @@ function getAllDatasets(vmobj)
...
+    topmost_dataset += '/';
+
     for (disk in vmobj.disks) {
+        var disk_dataset;
+
         disk = vmobj.disks[disk];
         if (disk.hasOwnProperty('zfs_filesystem')) {
-            datasets.push(disk.zfs_filesystem);
+            disk_dataset = disk.zfs_filesystem;
+            // Don't push if child of top-level zfs_filesystem.
```
bahamat, line 11653: Turn this comment around. Something like this:

```
// Child datasets of the zone root are automatically included because of -r (recursive).
// If the dataset name is not a child of the zone root, then add it to the list
// of datasets we need to transfer.
```

src/vm/node_modules/VM.js
```diff
@@ -11734,7 +11747,8 @@ function sendDataset(target, dataset, log, callback)
                     }
                 );
             }, function (cb) {
-                zfs(['snapshot', dataset + '@sending'], log,
+                // Use -r for the snapshot, in case of BHYVE datasets.
```
bahamat, line 11750: Instead:

```
// Use -r for the snapshot in case of any child datasets (bhyve, delegeated datasets, etc)
```

danmcd: Hadn't run delegated datasets when I wrote this comment.  Will fix.  Please mark this resolved when it is to your satisfaction.

src/vm/node_modules/VM.js
danmcd: Will fix.  Please mark this resolved when it is to your satisfaction.

bahamat (approved):

PR #1107, opened 2024-07-11
OS-8558 Pre-existing pools from foreign operating systems will break sysinfo

bahamat:


danmcd (approved):
IA will need some test notes about different kinds of failure here.

bahamat:
Testing notes in Jira.

PR #1109, opened 2024-09-12
TRITON-2453 sdc-fwrule parsing errors generate wrong error message

bahamat:


src/fw/node_modules/fwrule/index.js
```diff
@@ -192,14 +193,6 @@ parser.yy.parseError = function parseError(_, details) {
...
     }
 
-    if (details.text === '') {
-        err = new validators.InvalidParamError('rule',
-            'Error at character 0: \'\', expected: \'FROM\', found: '
-            + 'empty string');
-        err.details = details;
-        throw err;
-    }
-
```
danmcd, line 202: Same explanation (updated sdc- here as in sdc-fwrule#5, right?

bahamat: Yep, exactly the same. SmartOS just has its own full copy.

danmcd (approved):
Pardon latency...

danmcd:
Make sure smartos-live testing, if not already covered, is in the ticket with the other cases.

PR #1110, opened 2024-09-19
OS-8264 Pool-bootable SmartOS (and Triton) cannot handle sdc-factoryreset

danmcd:
OS-8578 Add "-s" option to sdc-factoryreset to poweroff CN instead of reboot

### DIFF
--- a/src/smartdc/bin/sdc-factoryreset.sh
+++ b/src/smartdc/bin/sdc-factoryreset.sh
@@ -7,6 +7,7 @@
 
 #
 # Copyright (c) 2014, Joyent, Inc.
+# Copyright 2024 MNX Cloud, Inc.
 #
 
 #
@@ -33,14 +34,26 @@ function usage()
 	exit 1
 }
 
+function wipe_bootpools()
+{
+	# disable booting on all pools
+	for pool in $(piadm bootable | grep -v non-bootable | awk '{print $1}')
+	do
+		piadm bootable -d $pool
+	done
+}
+
 if [[ -n $1 ]] && [[ $1 = "--help" ]]; then
 	usage
 fi
 
-while getopts "h" opt
+shutdown="0"
+
+while getopts "hs" opt
 do
 	case "$opt" in
 		h)	usage;;
+		s)	shutdown="1";;
 		*)	usage;;
 	esac
 done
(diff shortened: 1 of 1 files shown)


danmcd:
NOTE:  Needs testing, but conceptually this is good.  Will update ticket(s) with test results.

tsoome (commented):
You can do like this:

shutdown=""

in case statement
s) shutdown=yes;;

and then

if [[ $shutdown ]]; then
    poweroff
else
    reboot
fi

Might be just a bit easier to read (less chars to process;)

danmcd:
> You can do like this:
> 
> shutdown=""
> 
> in case statement s) shutdown=yes;;
> 
> and then
> 
> if [[ $shutdown ]]; then poweroff else reboot fi
> 
> Might be just a bit easier to read (less chars to process;)

Even better might be to just set a variable to the desired last-command!

nheimlich:
This may need to account for a shutdown rather than a reboot
https://github.com/TritonDataCenter/smartos-live/blob/4136f12f33b64e3324b5f118b0c69f6fc5462ec3/src/smartdc/bin/sdc-factoryreset.sh#L75

also.. usage nit
https://github.com/TritonDataCenter/smartos-live/blob/4136f12f33b64e3324b5f118b0c69f6fc5462ec3/src/smartdc/bin/sdc-factoryreset.sh#L31

nheimlich (approved):

src/smartdc/bin/sdc-factoryreset.sh
```diff
@@ -27,28 +28,41 @@ function abort()
...
 
+function wipe_bootpools()
+{
+	# disable booting on all pools -- XXX KEBE ASKS, make this optional?
+	for pool in $(piadm bootable | grep -v non-bootable | awk '{print $1}')
+	do
+		piadm bootable -d $pool
+	done
+}
+
```
bahamat, line 45: I'm very undecided as to whether I think this is ok here or if it should really be done in fs-joyent along with the rest of the pool resets.

danmcd: Me too, given work-Slack conversations (sorry folks in the open-source audience).

This is why 8264 was not examined a lot since its filing...  we may want to split out 8578, and maybe have "wipe bootable now" in `sdc-factoryreset` as its own non-default option as part of 8578, and fix 8264 in `fs-joyent`.

danmcd:
Moving this back to draft to cleave out 8578 as its own, lower-impact, PR, namely TritonDataCenter/smartos-live#1112

src/smartdc/bin/sdc-factoryreset.sh
danmcd: Much reduced now that 8578 is back.  Resolving this for now given if `fs-joyent` is the correct place for such fixes, we can eliminate this entirely.

PR #1111, opened 2024-10-31
OS-8586 illumos-joyent should use tools lex per illumos#16873

tsoome:
Once the lex update lands, we need to use it or gcc 14 build will fail:)


danmcd (approved):

danmcd:
The current `pr-head` build will be the final proof in the pudding.

PR #1112, opened 2024-11-19
OS-8578 Add "-s" option to sdc-factoryreset to poweroff CN instead of reboot

danmcd:
The lower-hanging, lesser-impact, change to `sdc-factoryreset` I'd like to make.

### DIFF
--- a/man/smartdc/man/man1/sdc-factoryreset.1.md
+++ b/man/smartdc/man/man1/sdc-factoryreset.1.md
@@ -3,23 +3,27 @@
 
 ## SYNOPSIS
 
-`sdc-factoryreset [-h | --help]`
+`sdc-factoryreset [-h | --help] {-s}`
 
 
 ## DESCRIPTION
 
 This command resets a machine to its originally installed state.  Specifically,
 it reboots the machine, imports all ZFS pools, and destroys them individually.
-It does this by setting a ZFS user property on the system pool.
+It does this by setting a ZFS user property on the system pool. Using the
+`-s` flag will instead shut down the machine; putting it into a state where
+upon next reboot will import all ZFS pool and destroy them individually.
 
 If this command is invoked unintentionally, an administrator can prevent the
-system from resetting itself by booting in rescue mode (noimport=true as a GRUB
-boot option) and clearing the smartdc:factoryreset property from the var
-dataset.  If the system is booting without the noimport=true GRUB option, the
-only way to stop the pending factory reset is to power cycle the machine, and
-boot again into rescue mode.  The service which does the actual factory reset
-starts well before an administrator would be able to login to the box, even if
-that administrator has console access.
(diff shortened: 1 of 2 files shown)


src/smartdc/bin/sdc-factoryreset.sh
bahamat, line 1: I think this is a better direction, but some of the prose still needs work. E.g., this will print out the following:

> WARNING: This machine will Rebooting and destroy its ZFS pools 
> after the next reboot.

or

> WARNING: This machine will Powering off and destroy its ZFS pools 
> after the next reboot.

It might be better to just use case statements.

bahamat (changes requested):

src/smartdc/bin/sdc-factoryreset.sh
danmcd: You're swapping `final_command` for `final_verb` I believe.  "This machine will reboot" or "This machine will poweroff" is what will get printed.

danmcd:
Testing notes in ticket.

man/smartdc/man/man1/sdc-factoryreset.1.md
```diff
@@ -12,14 +12,15 @@ This command resets a machine to its originally installed state.  Specifically,
...
+upon next reboot, the machine will import all ZFS pool and destroy them
+individually.
 
 If this command is invoked unintentionally, an administrator can prevent the
 system from resetting itself by booting in rescue mode (noimport=true as a
 loader boot option) and clearing the smartdc:factoryreset property from the
 var dataset.  Use of the `-s` option makes this easier to accomplish.
 
-If the affected system boots without the noimport=true GRUB option, the only
+If the affected system boots without the noimport=true loader option, the only
```
bahamat, line 23: Maybe just say boot option? That way it’s boot loader agnostic.

src/smartdc/bin/sdc-factoryreset.sh
```diff
@@ -37,18 +38,22 @@ if [[ -n $1 ]] && [[ $1 = "--help" ]]; then
...
 		*)	usage;;
 	esac
 done
 
 trap abort SIGINT
 
-printf "WARNING: This machine will reboot and destroy its ZFS pools after "
-printf "rebooting.\n"
+printf "WARNING: This machine will $final_command and destroy its ZFS pools "
+printf "after the next reboot.\n"
```
bahamat, line 56: Here, I would just say "after the next boot", which covers both boot from power off and reboot.

bahamat (dismissed):

man/smartdc/man/man1/sdc-factoryreset.1.md
danmcd: Will fix.

tsoome (approved):

bahamat (approved):

PR #1113, opened 2024-11-20
OS-8584 Introduce gcc14 as the shadow-compiler

tsoome:
Update Jenkinsfile for gcc14

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,7 @@
 
 /*
  * Copyright 2022 Joyent, Inc.
- * Copyright 2023 MNX Cloud, Inc.
+ * Copyright 2024 MNX Cloud, Inc.
  */
 
 @Library('jenkins-joylib@v1.0.8') _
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -255,7 +255,7 @@ export PLAT_CONFIGURE_ARGS="-d $PLAT_CONFIGURE_ARGS"
                     node {
```
danmcd, line 255: A couple lines above, you need to rename the stage to `gcc14` as well (line 253).

danmcd (changes requested):
Update copyright, and rename stage.  There should be 11 changed lines of gcc7 -> gcc14, plus a copyright update.

Jenkinsfile
tsoome: fixed.

danmcd (approved):
Will grant IA if the Jenkins build works!

Issue #1115, opened 2024-12-04
Docker images cannot be imported at all anymore.


Charadon:
Depending on the image trying to be imported, it either gives the error:

```
$ imgadm import docker.io/mageia:9
imgadm import: error (InvalidContent): no layers in mageia:9 manifest
```

or 

```
$ imgadm import docker.io/mageia:7
Uncaught TypeError: Cannot call method 'map' of undefined

FROM
/usr/img/lib/sources/docker.js:235:50
/usr/img/node_modules/docker-registry-client/lib/registry-client-v2.js:1351:9
next (/usr/img/node_modules/vasync/lib/vasync.js:201:4)
_afterCall (/usr/img/node_modules/docker-registry-client/lib/registry-client-v2.js:1347:17)
IncomingMessage.finish (/usr/img/node_modules/docker-registry-client/lib/docker-json-client.js:171:13)
IncomingMessage.g (events.js:180:16)
IncomingMessage.EventEmitter.emit (events.js:117:20)
_stream_readable.js:920:16
process._tickCallback (node.js:415:13)
Abort (core dumped)
```

I think it has something to do with the manifest api, but I don't know.

bahamat:
Yeah, dockerv2 registries are not supported. You can however, have your own v1 registry repo (or imgapi) that you import repos from docker hub into, then you can import them to SmartOS.

Proper dockerv2 support is only available in Triton via imgapi.

PR #1118, opened 2025-01-08
Support mediatype application/vnd.docker.distribution.manifest.list.v…

skippie81:
related to issues #1115 

recent docker images seem to return a manifest list for the /v2/<name>/manifests/<ref> call and not a manifest.
out of the manifest list this fix selects the digest for the manifest with platform architecture=amd64 and os=linux
and than this digest is used to call the manifest again to get the actual manifest with the image layers.

…2+json and application/vnd.oci.image.index.v1+json

For docker images support the manifest.list.v2 and oci.image.index.v1 when getting a manifest for a docker image in de registry-v2 client. When requesting a manifest a manifestList might be returned. In this case select the digest out of the manifestList for the correct platform.

platform needs to be:
  architecture: amd64
  os: linux

if manifest list contains a valid manifest use the digest to get the manifest with image layers.

### DIFF
--- a/src/img/lib/sources/docker.js
+++ b/src/img/lib/sources/docker.js
@@ -195,87 +195,97 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
     var importInfo;
 
     var client = self._clientFromRepo(rat);
-    client.getManifest({ref: rat.tag}, function (err, dockerManifest, res) {
-        if (err) {
-            if (err.statusCode === 404) {
-                cb();
-            } else {
-                cb(err);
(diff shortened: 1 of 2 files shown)


skippie81:
As requested. by @bahamat specific images to pull for testing

one that still works on current version (old image) returning a manifest by default: linuxserver/unifi:latest
one that is not working returning a manifestlist mediatype: imgadm import prom/alertmanager:v0.26.0
one that is not working returning a oci.imageindex mediatype: docker.io/linuxserver/unifi-network-application:latest

PR #1120, opened 2025-01-23
Makefile: Upgrade to Python 3

cclauss:
```diff
-			PYTHON="/opt/local/bin/python2.7" \
+			PYTHON="/opt/local/bin/python3" \
```
Python 2 died 1,849 days ago on 1/1/2020.
* Node.js no longer supports its use.
    * https://github.com/nodejs/node/issues/25789 was closed in February 2020.


bahamat:
Unfortunately it's not as simple as just replacing python2.7 with python3. There are well known breaking changes in python3 that need to be accounted for. We also use node v0.10 in the platform image which needs python2.7.

PR #1121, opened 2025-01-25
Modernize Python 2 code

cclauss:
Building a secure OS on insecure foundations is an increasingly difficult task.  `Makefile` currently builds with unsupported versions of both Node.js and Python.  This pull request focuses on modernizing the latter.

As discussed in #1120, the migration to Python 3 is a multistep process so these changes represent a safe [stage1](https://python-future.org/futurize.html#forwards-conversion-stage1).
* #1120
* https://portingguide.readthedocs.io
* https://docs.astral.sh/ruff/linter
* https://python-future.org

Find all files that contain Python 3 syntax errors.
% `ruff check --ignore=ALL --output-format=concise`

Modernize the code to continue to run on Python 2 but without any Python 3 syntax errors.
```
% uv tool run --python=3.12 --from=future futurize --stage1 --write \
    src/node-bunyan/tools/cutarelease.py \
    tools/javascriptlint/javascriptlint/conf.py \
    tools/javascriptlint/javascriptlint/jsl.py \
    tools/javascriptlint/javascriptlint/jsparse.py \
    tools/javascriptlint/javascriptlint/lint.py \
    tools/javascriptlint/javascriptlint/visitation.py \
    tools/javascriptlint/javascriptlint/warnings.py
```
% `ruff check --ignore=ALL --output-format=concise`
```
All checks passed!
```


bahamat:
Have you built a platform with this yet?

cclauss:
No.  I tried the Quickstart in README but I did not get too far.  On my MacOS box, I edited the `./configure` shebang from `/usr/bin/base` to `/bin/bash` but then I got `configure: build only works on SunOS`.  I was a Systems Engineer at Sun Microsystems for seven years around the turn of the century but these daze I have no access to a SunOS box.

My hope was that Jenkins would build the platform and raise issues where it was unable to do so.

bahamat:
Build instructions are in the README.

< https://github.com/TritonDataCenter/smartos-live>

You can run SmartOS in a VM to perform the build.

The reality is that the magnitude of the change you’re proposing is going to require a lot of work. Rather than attempt to pawn that work of on others, if you’d really like to see this change land, you’re going to put forth that effort yourself.

But a drive by pull request that hasn’t even been tested is not going to cut it.

cclauss:
Please stop assuming that I am lazy.  In #1120 I provided a link to all the steps in the long effort that I went through to successfully port Node.js (and upstream) from Python 2 to Python 3.  I could provide examples from other projects as well if that would be helpful but given that the first was not read, I doubt that would be the case.

The URL above returns a 404 and even when fixed does not provide someone on macOS enough information to successfully build the development environment.  Are any of this project's contributors developing on macOS?  I have lots of experience using Docker but it has been more than a decade since I used VMware.  Is Docker a possibility or is VMware required?  If so, is it required to download `VMware Fusion Pro for Mac` from Broadcom or is a different set of tools required?

bahamat:
In fact, all of the developers for SmartOS are using Macs, but the README very clearly states that the SmartOS build requires building on SmartOS.

I've corrected the link (my bad, I was on mobile) in my previous comment, but I'll put it again here as well: https://github.com/TritonDataCenter/smartos-live#building-smartos

It seems that you're not aware of what SmartOS even is. As its name implies, it's an Operating System and has nothing to do with Linux, so no, Docker won't work. It also runs only on x86, so you can use VMware Fusion if you're on a Mac with an Intel processor. If you have an Apple Silicon processor you can use UTM. Or can use a separate x86 host to run SmartOS directly, or as a VM.

There is extensive documentation for [installing SmartOS](https://docs.smartos.org/download-smartos/) with various installation targets. After which, you must read and carefully follow the [documentation for building SmartOS](https://github.com/TritonDataCenter/smartos-live#building-smartos) to set up your build environment before you can expect to make a successful build. I also highly recommend reading  the [Triton Contribution Guidelines](https://github.com/TritonDataCenter/smartos-live/blob/master/CONTRIBUTING.md) as well.

If you are having trouble with setting up a build environment, you can [reach out to the mailing list](https://docs.smartos.org/mailing-lists-and-irc/) since support requests are not handled through GitHub issues or PRs.

PR #1125, opened 2025-03-06
OS-8631 tests-tar target needs to handle symlinks better

danmcd:


danmcd:
Passed test it was supposed to (`secflags_dts`).

tsoome (approved):

nshalman (approved):

PR #1127, opened 2025-03-17
OS-8633: Want builder brand for RFD 184

nshalman:
https://smartos.org/bugview/OS-8633

We want `vmadm` to be able to create `builder` branded zones.
Depends on tritondatacenter/illumos-joyent#503

### DIFF
--- a/src/vm/etc/vmadm.completion
+++ b/src/vm/etc/vmadm.completion
@@ -5,7 +5,7 @@ _vmadm()
     # NOTE: we don't want the aliases boot,halt,destroy,etc here because we
     # want people to use the 'proper' commands when they can.
     COMMANDS="console create create-snapshot delete delete-snapshot get info list lookup reboot receive reprovision rollback-snapshot send start stop sysrq update validate"
-    BRANDS="bhyve joyent joyent-minimal kvm lx"
+    BRANDS="bhyve builder joyent joyent-minimal kvm lx"
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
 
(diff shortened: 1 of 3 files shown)


danmcd:
Same question about ADMINUI applies here... I should probably take a PI you've built with both this and the illumos-joyent changes and try it out.

danmcd (commented):
Can you confirm that you basically put "builder" whereever you say "joyent" ?  And you caught all of them?  Assuming yes to both, I'll +1 this.

danmcd (dismissed):
Thank you.  No IA until the whole wad (incl. adminui) is ready.

nshalman:
> Check if one isn't filed already, but we need to rewhack vmadm(8) to pass checks.
> 
> And at the risk of piling on, do we want to document builder(7)? What's frustrating is that joyent(7) is here, but lx(7) is in illumos-joyent. I think builder(7) should be here with joyent(7) OR joyent(7) gets moved to illumos-joyent.
> 
> No blockers at the moment, however.

https://smartos.org/bugview/OS-8645 filed for the man page line lengths

danmcd:
I also think, and I'll file this later today or tomorrow, we should:

1.) Decide if-or-not we move joyent(7) over to illumos-joyent.

2.) Regardless of 1's answer, either add sub-entries for joyent-minimal and builder and have them be symlinks to joyent(7), or have them get their own section-7 man pages.

nshalman:
> I also think, and I'll file this later today or tomorrow, we should:
> 
> 1.) Decide if-or-not we move joyent(7) over to illumos-joyent.
> 
> 2.) Regardless of 1's answer, either add sub-entries for joyent-minimal and builder and have them be symlinks to joyent(7), or have them get their own section-7 man pages.

Yes, the man page should live in the same repo as the brand.

danmcd (changes requested):
Fix copyright in proptable.js and I'll IA this.

danmcd (approved):

PR #1131, opened 2025-04-11
OS-8647 /lib/sdc/config.sh should validate the variable name

tsoome:
bash (and ksh) does not accept just any name for variable name.

### DIFF
--- a/src/lib/sdc/config.sh
+++ b/src/lib/sdc/config.sh
@@ -8,7 +8,7 @@
 # bash config.sh -json
 #
 # Copyright 2018 Joyent Inc.
-# Copyright 2023 MNX Cloud, Inc.
+# Copyright 2025 MNX Cloud, Inc.
 #
 
 CACHE_FILE_JSON="/tmp/.config.json"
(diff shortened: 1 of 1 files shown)


src/lib/sdc/config.sh
```diff
@@ -81,7 +81,11 @@ function load_sdc_sysinfo {
     /usr/bin/sysinfo -p | while read -r entry; do
         lval=$(echo $entry | cut -d= -f 1 | sed -e 's/#//g')
         rval=$(echo $entry | cut -d= -f 2-)
-        echo ${prefix}${lval}=${rval} >> $tmpfile
+
+	# lval must be identifier, that is sequence of letters, digits,
```
danmcd, line 85: Slight grammar and clarity fixes:
```
        # lval must be a bash identifier: a sequence of letters, digits,
        # or underscores
```

src/lib/sdc/config.sh
```diff
@@ -202,7 +206,10 @@ function load_sdc_bootparams {
     for line in $(/bin/bootparams); do
         fields=(${line//=/ })
         key=$(echo ${fields[0]} | sed -e "s/-/_/g;s/#//g")
-        eval "${prefix}${key}=\"${fields[1]}\""
+	# key must be identifier, that is sequence of letters, digits,
```
danmcd, line 209: Same sorta thing here:
```
        # key must be a bash identifier: a sequence of letters, digits,
        # or underscores
```

danmcd (changes requested):
Editorial nits and please update the MNX copyright to 2025.

danmcd (approved):
IA-ing too.

danmcd (approved):
Push it.

PR #1135, opened 2025-04-30
OS-8655 tests-tar tarball changes owner/permission on ./.

danmcd:
This popped up on IRC when someone was asking about our tests tarballs.  This will also require a documentation change in illumos-joyent/usr/src/test/smartos-tests/README as well.  Testing notes will be in the ticket shortly, as I did them once already.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,8 @@ $(TESTS_TARBALL): $(TESTS_MANIFEST)
 	cp $(STAMPFILE) $(ROOT)/tests.buildstamp
 	pfexec ./tools/builder/builder $(TESTS_MANIFEST) $(TESTS_PROTO) \
 	    $(PROTO) $(ROOT)
-	pfexec gtar -C $(TESTS_PROTO) -I pigz -cf $@ .
+	pfexec gtar -C $(TESTS_PROTO) -I pigz -cf $@ $(shell cd $(TESTS_PROTO) \
+		&& ls)
 	rm $(ROOT)/tests.buildstamp
 
 tests-tar: $(TESTS_TARBALL)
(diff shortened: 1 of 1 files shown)


tsoome (approved):

Makefile
```diff
@@ -289,7 +289,8 @@ $(TESTS_TARBALL): $(TESTS_MANIFEST)
 	cp $(STAMPFILE) $(ROOT)/tests.buildstamp
 	pfexec ./tools/builder/builder $(TESTS_MANIFEST) $(TESTS_PROTO) \
 	    $(PROTO) $(ROOT)
-	pfexec gtar -C $(TESTS_PROTO) -I pigz -cf $@ .
+	pfexec gtar -C $(TESTS_PROTO) -I pigz -cf $@ $(shell cd $(TESTS_PROTO) \
+		&& ls)
```
nshalman, line 293: In theory, if this tarball contained top level files or directories whose names started with `.` then those files would now be excluded. Am I correct? As long as it is never supposed to do that, I'm fine with this change.

nshalman (approved):
Approved as long as we know that this would exclude files that would be hidden from `ls`

Makefile
danmcd: Made it explicit in the comments now.

danmcd:
Testing notes in ticket still apply.

nshalman (approved):

cneira:
> Approved as long as we know that this would exclude files that would be hidden from `ls`

I second that, if the intention was that, and that was reflected in the last comments, this change looks good to me.

cneira (approved):

tsoome (approved):

PR #1139, opened 2025-06-04
OS-8653 PIADM(8) could be more verbose/helpful about curl(1) errors.

cneira:
The addition of the -S flag (--show-error) in verbose mode could be useful for troubleshooting.

### DIFF
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -330,15 +330,16 @@ piadm(8) -- Manage SmartOS Platform Images
 
     The following environment variables affect piadm(8) downloads internally.
 
-    PIADM_SUM_URL      If set, the value will be used to retrieve the checksums file, currently
-                            md5sums.txt file, which will used to validate the checksum of downloaded
-                            platform images.
+    PIADM_SUM_URL      If set, this properly-formed URL will be used to retrieve the checksums file.
+                            If not set, the default source of SmartOS PIs has a checksums file,
+                            based on PIADM_DIGEST_ALGORITHM (see below).
+                            The file is a list of {checksum, space, filename} tuples, one per line.
 
     PIADM_NO_SUM       If set to 1, this will skip checksum validation for downloaded platform
                             images.
 
-    PIADM_DIGEST_ALGORITHM  If set, the value will be used by digest(1) to calculate the checksums
-                            for downloaded platform images. By default md5 is used.
+    PIADM_DIGEST_ALGORITHM  If set, this value will be used by digest(1) to calculate checksums for
+                                 downloaded platform images (case sensitive). By default, md5 is used.
 
 ## EXIT STATUS
 
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -326,6 +326,16 @@ piadm(8) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+## ENVIRONMENT VARIABLES
+
+    The following environment variables affect the execution of PIADM(8).
+
+    PIADM_MD5SUM_URL    If set, the value will be used to retrieve the md5sums.txt file, which will
+                        used to validate the MD5 hashes of downloaded platform images.
+
+    PIADM_NO_MD5SUM     If set to 1, this will skip MD5 checksum validation for downloaded platform
+                        images.
+
 ## EXIT STATUS
 
 The following exit values are returned:
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -330,11 +330,14 @@ piadm(8) -- Manage SmartOS Platform Images
 
     The following environment variables affect the execution of PIADM(8).
 
-    PIADM_MD5SUM_URL    If set, the value will be used to retrieve the md5sums.txt file, which will
-                        used to validate the MD5 hashes of downloaded platform images.
+    PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the md5sums.txt file, which will
+                            used to validate the checksum of downloaded platform images.
 
-    PIADM_NO_MD5SUM     If set to 1, this will skip MD5 checksum validation for downloaded platform
-                        images.
+    PIADM_NO_CHECKSUM       If set to 1, this will skip checksum validation for downloaded platform
+                            images.
+
+    PIADM_DIGEST_ALGORITHM  If set, the value will be used by DIGEST(1) to calculate the checksums
+                            for downloaded platform images. By default md5 is used.
 
 ## EXIT STATUS
 
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -328,18 +328,18 @@ piadm(8) -- Manage SmartOS Platform Images
 
 ## ENVIRONMENT VARIABLES
 
-    The following environment variables affect piadm(8) downloads internally.
+The following environment variables affect piadm(8) downloads internally.
 
-    PIADM_SUM_URL      If set, this properly-formed URL will be used to retrieve the checksums file.
-                            If not set, the default source of SmartOS PIs has a checksums file,
-                            based on PIADM_DIGEST_ALGORITHM (see below).
-                            The file is a list of {checksum, space, filename} tuples, one per line.
+     PIADM_SUM_URL           If set, this properly-formed URL will be used to retrieve the checksums file.
+                             If not set, the default source of SmartOS PIs has a checksums file,
+                             based on PIADM_DIGEST_ALGORITHM (see below).
+                             The file is a list of {checksum, space, filename} tuples, one per line.
 
-    PIADM_NO_SUM       If set to 1, this will skip checksum validation for downloaded platform
-                            images.
+     PIADM_NO_SUM            If set to 1, this will skip checksum validation for downloaded platform
+                             images.
 
-    PIADM_DIGEST_ALGORITHM  If set, this value will be used by digest(1) to calculate checksums for
-                                downloaded platform images (case sensitive). By default, md5 is used.
+     PIADM_DIGEST_ALGORITHM  If set, this value will be used by digest(1) to calculate checksums for
+                             downloaded platform images (case sensitive). By default, md5 is used.
 
 ## EXIT STATUS
 
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -330,11 +330,11 @@ piadm(8) -- Manage SmartOS Platform Images
 
     The following environment variables affect piadm(8) downloads internally.
 
-    PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the checksums file, currently
+    PIADM_SUM_URL      If set, the value will be used to retrieve the checksums file, currently
                             md5sums.txt file, which will used to validate the checksum of downloaded
                             platform images.
 
-    PIADM_NO_CHECKSUM       If set to 1, this will skip checksum validation for downloaded platform
+    PIADM_NO_SUM       If set to 1, this will skip checksum validation for downloaded platform
                             images.
 
     PIADM_DIGEST_ALGORITHM  If set, the value will be used by DIGEST(1) to calculate the checksums
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -339,7 +339,7 @@ piadm(8) -- Manage SmartOS Platform Images
                             images.
 
     PIADM_DIGEST_ALGORITHM  If set, this value will be used by digest(1) to calculate checksums for
-                                 downloaded platform images (case sensitive). By default, md5 is used.
+                                downloaded platform images (case sensitive). By default, md5 is used.
 
 ## EXIT STATUS
 
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -337,7 +337,7 @@ piadm(8) -- Manage SmartOS Platform Images
     PIADM_NO_SUM       If set to 1, this will skip checksum validation for downloaded platform
                             images.
 
-    PIADM_DIGEST_ALGORITHM  If set, the value will be used by DIGEST(1) to calculate the checksums
+    PIADM_DIGEST_ALGORITHM  If set, the value will be used by digest(1) to calculate the checksums
                             for downloaded platform images. By default md5 is used.
 
 ## EXIT STATUS
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -328,7 +328,7 @@ piadm(8) -- Manage SmartOS Platform Images
 
 ## ENVIRONMENT VARIABLES
 
-    The following environment variables affect the execution of PIADM(8).
+    The following environment variables affect piadm(8) downloads internally.
 
     PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the md5sums.txt file, which will
                             used to validate the checksum of downloaded platform images.
--- a/man/usr/share/man/man8/piadm.8.md
+++ b/man/usr/share/man/man8/piadm.8.md
@@ -330,8 +330,9 @@ piadm(8) -- Manage SmartOS Platform Images
 
     The following environment variables affect piadm(8) downloads internally.
 
-    PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the md5sums.txt file, which will
-                            used to validate the checksum of downloaded platform images.
+    PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the checksums file, currently
+                            md5sums.txt file, which will used to validate the checksum of downloaded
+                            platform images.
 
     PIADM_NO_CHECKSUM       If set to 1, this will skip checksum validation for downloaded platform
                             images.
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -198,7 +198,7 @@ PIADM_CONF=/var/piadm/piadm.conf
 
 # fetch_csum
 #
-# Fetches a MD5 hash using a platform file as the key.
+# Fetches checksums using a platform file as the key.
 #
 # Arguments:
 #   $1 - URL from where to fetch the PI from. The name of the PI
@@ -230,7 +230,7 @@ fetch_csum() {
 			stamps_url="${URL_PREFIX%/}?limit=1024"
 			stamp=$("${CURL[@]}" "${stamps_url}" |\
 				json -ga -c "this.name.match(/Z$/)" | json -ag name |\
-				sort  | tail -1; exit ${PIPESTATUS[0]})
+				sort  | tail -1; exit "${PIPESTATUS[0]}")
 			code=$?
 			if [[ $code -ne 0 ]]; then
 				eecho "Curl failed fetching PI data from ${stamps_url}"
@@ -245,17 +245,18 @@ fetch_csum() {
 		fi
 		csum_platform=$("${CURL[@]}"  "${csum_url}" |\
 			awk -v pattern="${platform_file}"\
-			'$0 ~ pattern { print $1; exit}'; exit ${PIPESTATUS[0]})
+			'$0 ~ pattern { print $1; exit}'; exit "${PIPESTATUS[0]}")
 		code=$?
 		if [[ $code -ne 0 ]]; then
 			eecho "fetching checksums from ${csum_url}"
+			#force re-calculation of checksum on the next call.
+			csum_platform=""
 			return 1
 		fi
-		echo ${csum_platform}
+		echo "${csum_platform}"
 		return 0
 	else
-		eecho "not recalculating checksum for $1" &>2
-		echo ${csum_platform}
+		echo "${csum_platform}"
 		return 0
 	fi
 }
@@ -446,7 +447,7 @@ install() {
 			/bin/rm -rf "${tdir}"
 			fatal "Curl exit code $code"
 		fi
-		validate_csum $1 "${tdir}/smartos.iso"
+		validate_csum "$1" "${tdir}/smartos.iso"
 		code=$?
 		if [[ $code -ne 0 ]]; then
 			/bin/rm -rf "${tdir}"
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -178,8 +178,8 @@ piname_present_get_bootfs() {
 
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
-CURL=( curl -ks -f )
-VCURL=( curl -k -f --progress-bar --show-error)
+CURL=( curl -ks -f --show-error)
+VCURL=( curl -k -f --progress-bar)
 
 vcurl() {
 	if [[ $VERBOSE -eq 1 ]]; then
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -266,7 +266,7 @@ validate_md5sum() {
 		eecho "Could not get md5sum for PI code: ${code}"
 		return 1
 	fi
-	local_md5sum=$(md5sum "${2}" | cut -d' ' -f1; exit ${PIPESTATUS[0]})
+	local_md5sum=$(digest -a md5 "${2}")
 	code=$?
 	if [[ $code -ne 0 ]]; then
 		eecho "md5sum failed for $2"
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -178,7 +178,7 @@ piname_present_get_bootfs() {
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
 CURL=( curl -ks -f )
-VCURL=( curl -k -f --progress-bar )
+VCURL=( curl -k -f -S --progress-bar )
 
 vcurl() {
 	if [[ $VERBOSE -eq 1 ]]; then
@@ -397,7 +397,7 @@ install() {
 
 		# Do a URL reality check.
 		vecho "Attempting download of URL $1"
-		vcurl -o "${tdir}/download" "$1"
+		vcurl -o "${tdir}/download" "$1" 2>/dev/null || echo "$1 seems to be a boot stamp"
 		if [[ -e ${tdir}/download ]]; then
 			# Recurse with the downloaded file.
 			dload=$(mktemp)
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -290,7 +290,7 @@ validate_csum() {
 	if [[ "${published_csum}" != "${local_csum}" ]]; then
 		vecho "published_checksum: ${published_csum}"
 		vecho "local_checksum:     ${local_csum}"
-		eecho  "local file does not matches published checksum"
+		eecho  "local file does not match published checksum"
 		return 1
 	fi
 	return 0
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -202,8 +202,8 @@ PIADM_CONF=/var/piadm/piadm.conf
 #
 # Arguments:
 #   $1 - URL from where to fetch the PI from. The name of the PI
-#   		is expected to be the last path component of the URL,
-#   		or the literal string "latest".
+#   	 is expected to be the last path component of the URL,
+#   	 or the literal string "latest".
 #
 # Environment Variables:
 #   PIADM_NO_MD5SUM     - If set to 1, skips MD5 checksum validation.
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -196,23 +196,29 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
 # Default path for piadm's configuration
 PIADM_CONF=/var/piadm/piadm.conf
 
-
-# This will fetch a md5sum using platform file as key,
-# Argument $1 could be an URL, where to fetch the PI,
-# where we expect the name of the PI to be the last element from
-# the URL or the string latest, in that case we will need
-# to fetch the name of the PI for the latest release, and fetch
-# the correct md5sum from the page.
-# In the case of "latest" we limit the amount of PI names
-# from ${URL_PREFIX} to 1024, which is the max value allowed.
-# which will show us 1024 PIs, being the last one the latest.
-# We can disable the validation of the md5sum setting the environment
-# variable PIADM_NO_MD5SUM to 1, if you need to specify a custom PI URL then
-# the environment variable PIADM_MD5SUM_URL must be set to the location
-# of the md5sums.txt file retrieve the checksums from that location.
+# fetch_md5sum
+#
+# Fetches a MD5 hash using a platform file as the key.
+#
+# Arguments:
+#   $1 - URL from where to fetch the PI from. The name of the PI
+#   		is expected to be the last path component of the URL,
+#   		or the literal string "latest".
+#
+# Environment Variables:
+#   PIADM_NO_MD5SUM     - If set to 1, skips MD5 checksum validation.
+#   PIADM_MD5SUM_URL    - If set, overrides the default md5sums.txt URL
+#   					  and uses the value provided.
+#
+# Notes:
+#   - If "latest" is used, up to 1024 PIs will be listed from ${URL_PREFIX}.
+#     The last one is assumed to be the latest.
+#   - On error, this function exits with code 1, following PIADM(8)
+#     convention: the error indicates a failure, but no changes were made
+#     to the system.
 md5_platform=""
 fetch_md5sum() {
-	if [ -z "$md5_platform" ]; then
+	if [[ -z "$md5_platform" ]]; then
 		local platform_file
 		local stamp
 		if [ "$1" != "latest" ]; then
@@ -242,7 +248,7 @@ fetch_md5sum() {
 		code=$?
 		if [[ $code -ne 0 ]]; then
 			eecho "fetching md5sums from ${md5_url}"
-			return 2
+			return 1
 		fi
 		echo ${md5_platform}
 		return 0
@@ -253,8 +259,13 @@ fetch_md5sum() {
 	fi
 }
 
-# $1 is url for the object that was downloaded to fetch its md5sum
-# $2 is the local file to run md5sum to compare
+# validate_md5sum
+#
+# $1 is the URL from where the PI was downloaded.
+# $2 is the on disk PI image that was download from $1.
+#
+# Error code will always be 1, following the PIADM(8) convention where
+# a return code of 1 means: an error has occurred, but no change was made.
 validate_md5sum() {
 	if [[ $PIADM_NO_MD5SUM -eq 1 ]]; then
 		vecho "WARNING: Not checking md5sums"
@@ -270,13 +281,13 @@ validate_md5sum() {
 	code=$?
 	if [[ $code -ne 0 ]]; then
 		eecho "md5sum failed for $2"
-		return 2
+		return 1
 	fi
-	if [ "${published_md5sum}" != "${local_md5sum}" ]; then
+	if [[ "${published_md5sum}" != "${local_md5sum}" ]]; then
 		vecho "published_md5sum: ${published_md5sum}"
 		vecho "local_md5sum:     ${local_md5sum}"
 		eecho  "local file does not matches published md5sum"
-		return 3
+		return 1
 	fi
 	return 0
 }
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -196,7 +196,7 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
 # Default path for piadm's configuration
 PIADM_CONF=/var/piadm/piadm.conf
 
-# fetch_checksum
+# fetch_csum
 #
 # Fetches a MD5 hash using a platform file as the key.
 #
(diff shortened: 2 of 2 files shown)


src/piadm.sh
```diff
@@ -397,7 +397,7 @@ install() {
 
 		# Do a URL reality check.
 		vecho "Attempting download of URL $1"
-		vcurl -o "${tdir}/download" "$1"
+		vcurl -o "${tdir}/download" "$1" 2>/dev/null || echo "$1 seems to be a boot stamp"
```
danmcd: Linewrap with `\<nl>` please.

src/piadm.sh
```diff
@@ -178,7 +178,7 @@ piname_present_get_bootfs() {
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
 CURL=( curl -ks -f )
-VCURL=( curl -k -f --progress-bar )
+VCURL=( curl -k -f -S --progress-bar )
 
 vcurl() {
 	if [[ $VERBOSE -eq 1 ]]; then
```
danmcd, line 185: Below this there's a comment about verbose curls.  Please update it to reflect your new goodness here?

danmcd (changes requested):
Even on this initial review, there are more questions than answers.  We need a bit of a design overview.  -S only helps when nothing returns from the curl request.

I think I'm going to engage on the mailing list about this.  I worry about expectations both with and without -v.  ALSO, by default we use "-s" which literally means absolutely silent.

I wonder if we should exploit curl's MANY return codes for this?

src/piadm.sh
```diff
@@ -177,8 +177,10 @@ piname_present_get_bootfs() {
 
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
+# Including -S (--show-errors) with the intention of making curl's errors
+# easier to parse.
```
danmcd: I believe `-S` is a NOP if you don't have `-s` enabled.  I think we MIGHT need `-S` in the normal CURL case.  We want silence == success, but not failure.

An initial suggestion of:

```
CURL=( curl -ksfS )
VCURL=( curl -kf --progress-bar )
```
means even without "-v" we'll see failures of curl.  We do use vcurl() failure to move-on-and-try-something-else so in SOME cases we may need to be silent, though.  There are five callers of vcurl(), and you'll have to examine each one to see how it's (ab)used.

cneira:
> > Even on this initial review, there are more questions than answers. We need a bit of a design overview. -S only helps when nothing returns from the curl request.
> > I think I'm going to engage on the mailing list about this. I worry about expectations both with and without -v. ALSO, by default we use "-s" which literally means absolutely silent.
> > I wonder if we should exploit curl's MANY return codes for this?
> 
> Interesting, there's this in curl:
> 
> ```
>        -f, --fail
>               (HTTP) Fail with error code 22 and with no response body output
>               at all for HTTP transfers returning HTTP response codes at 400
>               or greater.
> 
>               In normal cases when an HTTP server fails to deliver a document,
>               it returns a body of text stating so (which often also describes
>               why and more) and a 4xx HTTP response code. This command line
>               option prevents curl from outputting that data and instead
>               returns error 22 early. By default, curl does not consider HTTP
>               response codes to indicate failure.
> 
>               To get both the error code and also save the content, use
>               --fail-with-body instead.
> 
>               This method is not fail-safe and there are occasions where
>               non-successful response codes slip through, especially when
>               authentication is involved (response codes 401 and 407).
> 
>               Providing --fail multiple times has no extra effect.  Disable it
>               again with --no-fail.
> ```
> 
> We can use that, probably in BOTH cases of vcurl(). I'll note that exit code 22 means (with -f on the command line) you got an http reporting code >= 400.

I'll take a look at that, I was thinking that maybe also we could leverage [-w](https://curl.se/docs/manpage.html#-w ) 
```
 -w, --write-out <format>
              Make curl display information on stdout after a completed
              transfer. The format is a string that may contain plain text
              mixed with any number of variables.  The format can be specified
              as a literal "string", or you can have curl read the format from
              a file with "@filename" and to tell curl to read the format from
              stdin you write "@-".

              The variables present in the output format are substituted by
              the value or text that curl thinks fit, as described below. All
              variables are specified as %{variable_name} and to output a
              normal % you just write them as %%. You can output a newline by
              using \n, a carriage return with \r and a tab space with \t.

```

src/piadm.sh
```diff
@@ -195,6 +196,91 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+		vecho "WARNING: Not checking md5sums"
+		return 0
+	fi
+	published_md5sum=$(fetch_md5sum $1)
+	code=$?
+	if [[ $code -ne 0 ]]; then
+		eecho "Could not get md5sum for PI code: ${code}"
+		return 1
+	fi
+	local_md5sum=$(md5sum "${2}" | cut -d' ' -f1; exit ${PIPESTATUS[0]})
```
danmcd: I prefer using always-on-platform `digest -a md5` for this, instead of a piece of software from pkgsrc.

This also simplifies the line above:
`local_md5sum=$(digest -a md5 "${2}")`

src/piadm.sh
```diff
@@ -195,6 +196,91 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+md5_platform=""
+fetch_md5sum() {
+	if [ -z "$md5_platform" ]; then
+		local platform_file
+		local stamp
+		if [ "$1" != "latest" ]; then
+			IFS=' ' read -r stamp platform_file <<<\
+			$(echo "$1" | rev | cut -d'/' -f1,2 | rev | tr -s "/" " ")
+		else
+			#latest is a special case, we need find out stamp
```
danmcd, line 229: Is this something we have somewhere else already and can be factored out?

danmcd (approved):
Some initial comments...

danmcd (commented):
Did not mean to approve.

danmcd (changes requested):
Damn... let's try this again.

src/piadm.sh
cneira: I removed the flag and added checksum validation for downloaded PIs.

src/piadm.sh
cneira: Thank you, good catch!.

src/piadm.sh
cneira: When we have the stamp of the PI, we can forge the URL from where to fetch the ISO. This is not the case when  only the URL for the PI is provided. In this case we extract the stamp and file name from the URL path in order to query the md5sums.txt published with each release, because of that deconstructing the URL was required.

src/piadm.sh
```diff
@@ -13,10 +13,11 @@
 
 #
 # Copyright 2022 Joyent, Inc.
-# Copyright 2024 MNX Cloud, Inc.
+# Copyright 2025 MNX Cloud, Inc.
 #
 
 # shellcheck disable=1091
+# shellcheck disable=SC2317
```
danmcd: My shellcheck-fu is weak, why is this here?

(Looks it up...) Was this shellcheck error around BEFORE these changes?
- If so, don't worry about it right now but do file a followup bug to remove the shellcheck noise and this disable line.
- If not, please fix new shellcheck noise in this changeset unless you have a good reason I can't obviously see.

src/piadm.sh
```diff
@@ -195,6 +196,91 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+# variable PIADM_NO_MD5SUM to 1, if you need to specify a custom PI URL then
+# the environment variable PIADM_MD5SUM_URL must be set to the location
+# of the md5sums.txt file retrieve the checksums from that location.
+md5_platform=""
+fetch_md5sum() {
+	if [ -z "$md5_platform" ]; then
+		local platform_file
+		local stamp
+		if [ "$1" != "latest" ]; then
+			IFS=' ' read -r stamp platform_file <<<\
```
danmcd: The point is to extract the build stamp here for precision seeking on Manta. Currently we extract the stamp from the PI tarball or mounted ISO.  IF we extract the stamp here, might we be able to keep it for redundancy, etc.?

Also, can you shrink this by using `IFS='/'` instead? Or am I missing some subtlety?

man/usr/share/man/man8/piadm.8.md
```diff
@@ -326,6 +326,16 @@ piadm(8) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+## ENVIRONMENT VARIABLES
+
+    The following environment variables affect the execution of PIADM(8).
```
danmcd: 1.) Lowercase to `piadm(8)`.

2.) I'd HIGHLY recommend using `SUM` instead of `MD5SUM` for the variable names in case we opt to change MD5 out later. I'll talk about this a bit more in the code itself, but this place is OBVIOUS to not commit to a specific algorithm.

3.) Add an explicit sentence or two.  Here's my suggestion (ignore/edit/use as you see fit):

`These variables only affect images piadm(8) downloads internally.`

4.) The code may make it more clear, but if I utter:

`piadm install https://kebe.com/~danmcd/webrevs/platform-OS-xxxx.tgz`

will I have to use `PIADM_NO_SUM` (assuming you take my advice in 2) to not have it go searching for an md5sums.txt file?

man/usr/share/man/man8/piadm.8.md
```diff
@@ -326,6 +326,16 @@ piadm(8) -- Manage SmartOS Platform Images
...
+
+    PIADM_MD5SUM_URL    If set, the value will be used to retrieve the md5sums.txt file, which will
```
danmcd: "md5sums.txt" ==> "checksums file, currently `md5sums.txt`".

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+#   PIADM_MD5SUM_URL    - If set, overrides the default md5sums.txt URL
+#   					  and uses the value provided.
+#
+# Notes:
+#   - If "latest" is used, up to 1024 PIs will be listed from ${URL_PREFIX}.
+#     The last one is assumed to be the latest.
+#   - On error, this function exits with code 1, following PIADM(8)
+#     convention: the error indicates a failure, but no changes were made
+#     to the system.
+md5_platform=""
```
danmcd: Genericize, please.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+			json -ga -c "this.name.match(/Z$/)" | json -ag name |\
+				sort  | tail -1; exit ${PIPESTATUS[0]})
+			code=$?
+			if [[ $code -ne 0 ]]; then
+				eecho "Curl failed fetching PI data from ${stamps_url}"
+				return 1
+			fi
+			platform_file="smartos-${stamp}.iso"
+		fi
+		if [[ -z ${PIADM_MD5SUM_URL} ]];then
```
danmcd: Genercize `md5_url` to `csum_url` or something?

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
 # Default path for piadm's configuration
 PIADM_CONF=/var/piadm/piadm.conf
 
+# fetch_md5sum
```
danmcd: As I mentioned earlier, we *do* use MD5 all over for this right now.  On the off chance we move to some other hash (e.g. SHA256) in the future, do we want to be algorithm-agile here?

I think the function name (fetch_md5sum) should be independent (fetch_sum).  I think the environment names should be independent (see above).  I think internal variable names should be independent too.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+		echo ${md5_platform}
+		return 0
+	else
+		eecho "not recalculating md5sum for $1" &>2
+		echo ${md5_platform}
+		return 0
+	fi
+}
+
+# validate_md5sum
```
danmcd: validate_sum or validate_csum

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+				return 1
+			fi
+			platform_file="smartos-${stamp}.iso"
+		fi
+		if [[ -z ${PIADM_MD5SUM_URL} ]];then
+			local md5_url="${URL_PREFIX}${stamp}/md5sums.txt"
+		else
+			local md5_url="${PIADM_MD5SUM_URL}"
+		fi
+		md5_platform=$("${CURL[@]}"  "${md5_url}" |\
```
danmcd: This should be able to be de-MD5-named, I think.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+			awk -v pattern="${platform_file}"\
+			'$0 ~ pattern { print $1; exit}'; exit ${PIPESTATUS[0]})
+		code=$?
+		if [[ $code -ne 0 ]]; then
+			eecho "fetching md5sums from ${md5_url}"
```
danmcd: You can mention md5 here, but as a separate thing:  "fetching md5 sums..." for when we wanna push further into algorithm-independence.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+			return 1
+		fi
+		echo ${md5_platform}
+		return 0
+	else
+		eecho "not recalculating md5sum for $1" &>2
```
danmcd: Same here

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+		vecho "WARNING: Not checking md5sums"
+		return 0
+	fi
+	published_md5sum=$(fetch_md5sum $1)
+	code=$?
+	if [[ $code -ne 0 ]]; then
+		eecho "Could not get md5sum for PI code: ${code}"
+		return 1
+	fi
+	local_md5sum=$(digest -a md5 "${2}")
```
danmcd: Keep `digest -a md5` for now, but change variable name to local_sum/local_csum.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+	code=$?
+	if [[ $code -ne 0 ]]; then
+		eecho "md5sum failed for $2"
```
danmcd: See earlier.

src/piadm.sh
```diff
@@ -195,6 +196,102 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+# $2 is the on disk PI image that was download from $1.
+#
+# Error code will always be 1, following the PIADM(8) convention where
+# a return code of 1 means: an error has occurred, but no change was made.
+validate_md5sum() {
+	if [[ $PIADM_NO_MD5SUM -eq 1 ]]; then
+		vecho "WARNING: Not checking md5sums"
+		return 0
+	fi
+	published_md5sum=$(fetch_md5sum $1)
```
danmcd: published_sum or published_csum

src/piadm.sh
```diff
@@ -346,6 +443,12 @@ install() {
 			/bin/rm -rf "${tdir}"
 			fatal "Curl exit code $code"
 		fi
+		validate_md5sum $1 "${tdir}/smartos.iso"
```
danmcd: validate_sum/validate_csum

src/piadm.sh
```diff
@@ -346,6 +443,12 @@ install() {
...
+		code=$?
+		if [[ $code -ne 0 ]]; then
+			/bin/rm -rf "${tdir}"
+			err "validate_md5sum exit code  $code"
```
danmcd: validate_sum/validate_csum

src/piadm.sh
```diff
@@ -403,7 +506,11 @@ install() {
 			dload=$(mktemp)
 			mv -f "${tdir}/download" "$dload"
 			/bin/rm -rf "${tdir}"
-
+			validate_md5sum  "$1" "$dload"
```
danmcd: validate_sum/validate_csum

man/usr/share/man/man8/piadm.8.md
cneira: > 4.) The code may make it more clear, but if I utter:
> 
> `piadm install https://kebe.com/~danmcd/webrevs/platform-OS-xxxx.tgz`
> 
> will I have to use `PIADM_NO_SUM` (assuming you take my advice in 2) to not have it go searching for an md5sums.txt file?

This case will fail. PIADM(8) will use the default URL to fetch the md5sums.txt file (https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/<PI STAMP>/md5sums.txt).  But that's not the only problem.
The URL https://kebe.com/~danmcd/webrevs/platform-OS-xxxx.tgz does not contain the stamp value in the URL path; we currently rely on this to retrieve the checksum for the file. If the user sets PIADM_NO_SUM=1 and uses this URL to download a PI, then it will download the PI but not checksum it. 

For a custom URL to work with checksum validation, it would need to comply with the following : 
  
1. The stamp value of the PI needs to be placed in the URL segment before the segment that contains the file name of the PI. The current assumption is that the file name value for the PI is in the last segment of the URL. 
For example:  https://kebe.com/~danmcd/webrevs/20250529T000820Z/platform-OS-20250529T000820Z.tgz

2. The environment variable PIADM_SUM_URL should be set to the location of the checksum file for the PI.

3. The checksum file is a text file containing the checksum value and file names separated by a space. 
     For example : 
```
2a2f0916cdbb9429abeb8cf57e43375b platform-OS-20250529T000820Z.tgz
```
I'm thinking that we should set PIADM_NO_SUM=1 in the case that a URL is not from a known domain. 
If the user requires checksum validation, then PIADM_SUM_URL should be set by the user to the required value.

man/usr/share/man/man8/piadm.8.md
```diff
@@ -326,6 +326,19 @@ piadm(8) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+## ENVIRONMENT VARIABLES
+
+    The following environment variables affect the execution of PIADM(8).
+
+    PIADM_CHECKSUM_URL      If set, the value will be used to retrieve the md5sums.txt file, which will
```
danmcd: NIT.  If/when we allow other hash algorithms, `md5sums.txt` is probably not gonna be the name.

man/usr/share/man/man8/piadm.8.md
danmcd: That answers 4.

1 hasn't been fixed yet, neither has 3.

2 is moving along nicely, modulo some nits I might find.

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+# Fetches a MD5 hash using a platform file as the key.
+#
+# Arguments:
+#   $1 - URL from where to fetch the PI from. The name of the PI
+#   	 is expected to be the last path component of the URL,
+#   	 or the literal string "latest".
+#
+# Environment Variables:
+#   PIADM_NO_CHECKSUM     - If set to 1, skips checksum validation.
+#   PIADM_CHECKSUM_URL    - If set, overrides the default md5sums.txt URL
```
danmcd: Same advice per the man page about `md5sums.txt`.  It's the current default value, and the only one available.

src/piadm.sh
danmcd: Didn't get an answer for this. There's probably a good reason I'm missing for you using the bigger pipeline instead of `IFS='/'`.

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+				sort  | tail -1; exit ${PIPESTATUS[0]})
+			code=$?
+			if [[ $code -ne 0 ]]; then
+				eecho "Curl failed fetching PI data from ${stamps_url}"
+				return 1
+			fi
+			platform_file="smartos-${stamp}.iso"
+		fi
+		if [[ -z ${PIADM_CHECKSUM_URL} ]];then
+			local checksum_url="${URL_PREFIX}${stamp}/md5sums.txt"
```
danmcd: Yep, THIS is where we keep things hardcoded for now.  Good job! (Can resolve this upon reading.)

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+# $1 is the URL from where the PI was downloaded.
+# $2 is the on disk PI image that was download from $1.
+#
+# Error code will always be 1, following the PIADM(8) convention where
+# a return code of 1 means: an error has occurred, but no change was made.
+# Environment variable PIADM_DIGEST_ALGORITHM controls the checksum 
+# algorithm used by DIGEST(1), by default md5 is used.  
+validate_checksum() {
+	if [[ $PIADM_NO_CHECKSUM -eq 1 ]]; then
+		vecho "WARNING: Not validation checksum"
```
danmcd: "Not using validation checksum" is clearer.

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+		return 0
+	fi
+	published_md5sum=$(fetch_checksum "$1")
```
danmcd: We might wanna change these variables to not include md5 in their names.

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+	code=$?
+	if [[ $code -ne 0 ]]; then
+		eecho "Could not get checksum for PI code: ${code}"
+		return 1
+	fi
+	local_md5sum=$(digest -a "${PIADM_DIGEST_ALGORITHM-md5}" "${2}")
```
danmcd: Yeah, especially since you can override it now.

danmcd: I'm tempted to include in this release as supplemental material `sha256sums.txt` to help you test this.

danmcd: (I'll have to hand-craft it for now, but eventually the jenkins build will do it automatically for at least one other algorithm.)

danmcd (commented):
Good progress, but some things are incomplete.

src/piadm.sh
cneira: There was no justified reason for that pipeline. I overlooked simplifying it; it was a remnant of old code I used before IFS to simplify the original statement, except that I missed factoring out the rest of the line you identified.

src/piadm.sh
cneira: I removed the disabling of shellcheck warning id=SC2317 and fixed shellcheck noise that this change introduced.

src/piadm.sh
cneira: Thank you!.

src/piadm.sh
```diff
@@ -195,6 +196,106 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+
+# validate_csum
+#
+# $1 is the URL from where the PI was downloaded.
+# $2 is the on disk PI image that was download from $1.
+#
+# Error code will always be 1, following the PIADM(8) convention where
+# a return code of 1 means: an error has occurred, but no change was made.
+# Environment variable PIADM_DIGEST_ALGORITHM controls the checksum
+# algorithm used by DIGEST(1), by default md5 is used.
```
danmcd: You fixed it with `piadm`(8) in the man page, but digest(1) needs to be lowercase.

src/piadm.sh
```diff
@@ -195,6 +196,106 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+				sort  | tail -1; exit "${PIPESTATUS[0]}")
+			code=$?
+			if [[ $code -ne 0 ]]; then
+				eecho "Curl failed fetching PI data from ${stamps_url}"
+				return 1
+			fi
+			platform_file="smartos-${stamp}.iso"
+		fi
+		if [[ -z ${PIADM_SUM_URL} ]];then
+			local csum_url="${URL_PREFIX}${stamp}/md5sums.txt"
```
danmcd: Since you have all the other pieces in place:  ${PIADM_DIGEST_ALGORITHM-md5}sums.txt` ?

src/piadm.sh
danmcd: See my one last new comment, but I'm closing this since hand-crafting one will be helpful.

src/piadm.sh
danmcd: Seems like you did most of this save one new comment's worth.

src/piadm.sh
danmcd: Cool!

src/piadm.sh
danmcd: Cool, thanks.

danmcd (commented):
Almost there, two small ones left, I *think*.

src/piadm.sh
danmcd: This one still needs fixing.

cneira:
> Does this current batch of changes address the:
> 
> `piadm install https://kebe.com/~danmcd/work/try-this-PI.tgz`
> 
> problem where there's no downloadable checksum available?

Yes, that's right. This will disable checksum validation for explicit URLs, I'll update the ticket with the testing done on this set of changes.

src/piadm.sh
```diff
@@ -195,6 +196,107 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+#
+# Environment Variables:
+#   PIADM_NO_SUM     - If set to 1, skips checksum validation.
+#   PIADM_SUM_URL    - If set, overrides the default URL used
+#                      the checksum file used for platform images.
+#
+# Notes:
+#   - If "latest" is used, up to 1024 PIs will be listed from ${URL_PREFIX}.
+#     The last one is assumed to be the latest.
+#   - On error, this function exits with code 1, following PIADM(8)
```
danmcd: `piadm`.  Commands should be lowercased.

src/piadm.sh
```diff
@@ -195,6 +196,107 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+			'$0 ~ pattern { print $1; exit}'; exit "${PIPESTATUS[0]}")
+		code=$?
+		if [[ $code -ne 0 ]]; then
+			eecho "fetching checksums from ${csum_url}"
+			#force re-calculation of checksum on the next call.
+			csum_platform=""
+			stamp=""
+			return 1
+		fi
+		echo "${csum_platform}"
```
danmcd: You could have this fallthrough:

```
          ...
          return 1
      fi
      # Else fallthrough...
  fi
  echo "${csum_platform}"
  return 0
}

File: src/piadm.sh [hunk 33]
```diff
@@ -195,6 +196,107 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+	fi
+	local_csum=$(digest -a "${PIADM_DIGEST_ALGORITHM:-md5}" "${2}")
+	code=$?
+	if [[ $code -ne 0 ]]; then
+		eecho "checksum failed for $2, algorithm used: ${PIADM_DIGEST_ALGORITHM}"
+		return 1
+	fi
+	if [[ "${published_csum}" != "${local_csum}" ]]; then
+		vecho "published_checksum: ${published_csum}"
+		vecho "local_checksum:     ${local_csum}"
```
danmcd: I'd move these vecho OUTSIDE the inequality check before it.  If I'm using -v I wanna see these anyway.

src/piadm.sh
```diff
@@ -346,10 +448,18 @@ install() {
 			/bin/rm -rf "${tdir}"
 			fatal "Curl exit code $code"
 		fi
+		validate_csum "$1" "${tdir}/smartos.iso"
+		code=$?
+		if [[ $code -ne 0 ]]; then
+			/bin/rm -rf "${tdir}"
+			err "validate_csum exit code  $code"
```
danmcd: This is vague output.  Be explicit:
```
    err "Cannot validate checksum for $1"
```

danmcd (commented):
Now that the nits are out of the way, I'm stepping back to examine the big picture.

- Did you drop the changes to literally fix the stated problem ("curl errors")?  I thought we needed to fix the vcurl() function a bit? We need to make sure we're solving the initially stated problem fully (e.g. reporting a 503).

- For downloaded artifacts, are we only doing the checksum check for ISOs?  There's no wrong answer here, but the answer to this question needs to be documented near/at the new environment variables.

src/piadm.sh
```diff
@@ -179,7 +179,7 @@ piname_present_get_bootfs() {
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
 CURL=( curl -ks -f )
```
danmcd: I think you either need -S/--show-error HERE in regular $CURL, OR (see below)...

src/piadm.sh
```diff
@@ -514,20 +511,29 @@ install() {
...
-			return 0
+		# Check if URL exists first
+		vecho "Checking if URL $1 exists"
+		if ! "${CURL[@]}" --max-time 30 --connect-timeout \
+				10 --head "$1" > /dev/null 2>&1; then
+			vecho "URL $1 is not accessible or does not exist"
+			# Fall through to treat as boot stamp
+		else
+			vecho "Downloading from URL $1"
+			if vcurl -o "${tdir}/download" "$1"; then
```
danmcd, line 541: ... HERE, add "-S" to this invocation.

danmcd (commented):
Better, but missing one thing...

src/piadm.sh
```diff
@@ -512,10 +512,19 @@ install() {
 		# Explicit boot stamp or URL.
 
 		# Check if URL exists first
+		# XXX I think max-time 30s to fetch headers
```
danmcd: Swap out "XXX I think" for "We believe".

src/piadm.sh
```diff
@@ -512,10 +512,19 @@ install() {
...
+		# is more than enough.
 		vecho "Checking if URL $1 exists"
```
danmcd: Additionally, add " (30s timeout)" to the end of this `vecho` string.

src/piadm.sh
```diff
@@ -178,7 +178,7 @@ piname_present_get_bootfs() {
 
 # Defined as a variable in case we need to add parameters (like -s) to it.
 # WARNING:  Including -k for now.
-CURL=( curl -ks -f )
+CURL=( curl -ks -f --show-error)
```
danmcd, line 181: THANK YOU!

src/piadm.sh
```diff
@@ -512,10 +512,19 @@ install() {
...
 
 		# Check if URL exists first
+		# XXX I think max-time 30s to fetch headers
+		# is more than enough.
 		vecho "Checking if URL $1 exists"
 		if ! "${CURL[@]}" --max-time 30 --connect-timeout \
 				10 --head "$1" > /dev/null 2>&1; then
-			vecho "URL $1 is not accessible or does not exist"
+			# Get HTTP status code for error reporting
+			http_code=$(curl -ks -w "%{http_code}" --max-time 30 --connect-timeout \
```
danmcd: This feels like one-too-many curl invocations...  when I big-picture review this again later today I'll figure out more.

danmcd (commented):
This is a small batch on the most recent incremental changes.  I definitely need to pull back and re-review the whole wad.

man/usr/share/man/man8/piadm.8.md
```diff
@@ -326,6 +326,20 @@ piadm(8) -- Manage SmartOS Platform Images
  [root@smartos ~]#
 ```
 
+## ENVIRONMENT VARIABLES
+
+    The following environment variables affect piadm(8) downloads internally.
+
+    PIADM_SUM_URL      If set, the value will be used to retrieve the checksums file, currently
+                            md5sums.txt file, which will used to validate the checksum of downloaded
```
danmcd: Upon further reading, this needs to be called out as an ACTUAL FULL URL.  The default, as stated is a combination of: `${URL_PREFIX}${stamp}/${PIADM_DIGEST_ALGORITHM`.  Make sure that the default language is clearer too.  Lemme try a replacement section:
```
PIADM_SUM_URL  If set, this properly-formed URL will be used to retrieve the checksums file. If not set, the default source of SmartOS PIs has a checksums file, based on PIADM_DIGEST_ALGORITHM (see below). The file is a list of {checksum, space, filename} tuples, one per line.
```

man/usr/share/man/man8/piadm.8.md
```diff
@@ -326,6 +326,20 @@ piadm(8) -- Manage SmartOS Platform Images
...
+                            platform images.
+
+    PIADM_NO_SUM       If set to 1, this will skip checksum validation for downloaded platform
+                            images.
+
+    PIADM_DIGEST_ALGORITHM  If set, the value will be used by digest(1) to calculate the checksums
```
danmcd: Note that it is case-sensitive.

src/piadm.sh
danmcd: Make sure your (exhaustive so far, NICE!) test notes cover this change.

src/piadm.sh
danmcd: I'm imagining one less curl invocation (and my shell-fu may be weak and missing something...):

```
    if ! "${CURL[@]" --max-time 30 --connect-timeout \
        10 --head "$1" -w "%{http_code}" 2>/dev/null | tail -1 > /var/run/code.$$; then
        http_code=$(cat /var/run/code.$$)
        /bin/rm -f /var/run/code.$$
        ...
```

src/piadm.sh
danmcd: You did in vcurl.

danmcd (commented):
You're really close.  I'm sorry this has been such a slog, but we wanna get it right AND make sure we cover the impetus behind this in the first place.

Remaining global questions:

1. Can the three new variables get read from `/var/piadm/piadm.conf` ? Right now that's an empty file, but in the past it had URL prefixes (`URL_PREFIX`) during the joyent => MNX move.

2. The fewer curl invocations the better

3. There's one more possible optimization I can think of, but I'm not holding THIS push up, and it can be inspected concurrently with a "move to sha256" future RFE for smartos-live builds.

src/piadm.sh
```diff
@@ -512,14 +515,15 @@ install() {
...
+		vecho "Checking if URL $1 exists (30s) timeout"
 		if ! "${CURL[@]}" --max-time 30 --connect-timeout \
-				10 --head "$1" > /dev/null 2>&1; then
+				10 --head "$1" -w "%{http_code}" 2> /dev/null |\
+				tail -1 > /var/run/piadm.http.$$; then
 			# Get HTTP status code for error reporting
-			http_code=$(curl -ks -w "%{http_code}" --max-time 30 --connect-timeout \
-				10 --head -o /dev/null "$1" 2>/dev/null)
+			http_code=$(cat /var/run/piadm.$$)
+			/bin/rm -f /var/run/piadm.http.$$
```
danmcd, line 533: You leak this file if the check succeeds too.  Remove it in the `else` branch as well, please.

src/piadm.sh
```diff
@@ -355,6 +357,7 @@ EOF
 
 	# Can furthermore be overridden by the user's PIADM_URL_PREFIX.
 	URL_PREFIX=${PIADM_URL_PREFIX:-${DEFAULT_URL_PREFIX}}
+	source "${PIADM_CONF}"
```
danmcd, line 367: EDITED:  I think I see what's going on.

src/piadm.sh
```diff
@@ -1660,6 +1664,7 @@ bootparams | grep -q 'headnode=' && initialize_as_HN
 # Check the configuration file out.
 config_check
 
+
```
danmcd: Lose the whitespace.

danmcd (commented):
Be careful, you missed two corner cases.

src/piadm.sh
danmcd: Still not fixed!

cneira: I'm sorry it's fixed now!.

danmcd (dismissed):
If someone edits /var/piadm/piadm.conf they run a risk, but honestly I think that's a valid risk at this point.

I will IA after reviewing the testing notes again on the ticket.

cneira:
> If someone edits /var/piadm/piadm.conf they run a risk, but honestly I think that's a valid risk at this point.
> 
> I will IA after reviewing the testing notes again on the ticket.

Awesome!. Thank you for checking this out.

src/piadm.sh
```diff
@@ -195,6 +196,104 @@ DEFAULT_URL_PREFIX=https://us-central.manta.mnx.io/Joyent_Dev/public/SmartOS/
...
+#     The last one is assumed to be the latest.
+#   - On error, this function exits with code 1, following piadm(8)
+#     convention: the error indicates a failure, but no changes were made
+#     to the system.
+declare csum_platform=""
+declare stamp=""
+fetch_csum() {
+	if [[ -z "$csum_platform" ]]; then
+		local platform_file
+		if [ "$1" != "latest" ]; then
```
airencracken: Why not use `[[` here as well for consistency?

cneira: Thank you for catching this. I just fixed it.

danmcd (approved):
Good catch @airencracken .

danmcd:
@airencracken mentioned that his nit was fixed, but didn't want to officially approve.  I'm IA-ing this. Please use the edgecast email:
```
Reviewed by: Dan McDonald <danmcd@edgecast.io>
Approved by: Dan McDonald <danmcd@edgecast.io>
```

PR #1141, opened 2025-06-26
OS-8645 vmadm.8 man page has lines longer than 80 columns

cneira:
In this PR, I added Makefile targets to check for lines longer than 80 columns in every manpage that we have in smartos-live.
As a result, I fixed the man pages that failed these checks. I'm not sure about if this the best approach, but as today, [this check is already present in fw](https://github.com/TritonDataCenter/smartos-live/blob/master/src/fw/Makefile#L101)  so I think it won't hurt to make it available to the rest of man pages.
I need to inspect more closely the man pages that I have modified, so I'm marking this PR as draft, but I wanted to open the discussion.

### DIFF
--- a/man/Makefile
+++ b/man/Makefile
@@ -1,6 +1,6 @@
 #
 # Copyright 2020 Joyent, Inc.
-# Copyright 2022 MNX Cloud, Inc.
+# Copyright 2025 MNX Cloud, Inc.
 #
 
 ROOT = $(PWD)/..
--- a/src/Makefile
+++ b/src/Makefile
@@ -11,6 +11,7 @@
 
 #
 # Copyright 2019 Joyent, Inc.
+# Copyright 2025 MNX Cloud, Inc.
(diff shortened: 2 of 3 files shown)


man/Makefile
```diff
@@ -1,6 +1,6 @@
 #
 # Copyright 2020 Joyent, Inc.
-# Copyright 2022 MNX Cloud, Inc.
+# Copyright 2025 MNX Cloud, Inc.
```
danmcd, line 3: Bump this up to Edgecast, please?

src/Makefile
```diff
@@ -11,6 +11,7 @@
 
 #
 # Copyright 2019 Joyent, Inc.
+# Copyright 2025 MNX Cloud, Inc.
```
danmcd, line 14: Same here.

danmcd (commented):
Copyright nits.

PR #1143, opened 2025-07-01
OS-8675 imgadm segfaults importing docker image from V2 repository

cneira:
Testing notes in the ticket.

### DIFF
--- a/src/img/lib/sources/docker.js
+++ b/src/img/lib/sources/docker.js
@@ -21,9 +21,8 @@
  * CDDL HEADER END
  *
  * Copyright (c) 2018, Joyent, Inc. All rights reserved.
+ * Copyright 2025 Edgecast Cloud LLC.
  */
-
-var p = console.log;
 
 var assert = require('assert-plus');
 var concat = require('concat-stream');
@@ -195,7 +194,13 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
     var importInfo;
 
     var client = self._clientFromRepo(rat);
-    client.getManifest({ref: rat.tag}, function (err, dockerManifest, res) {
+    client.getManifestRef({ref: rat.tag}, function(err_, dockerManifestRef, res) {
+        if (err_ && err_.statusCode !== 404) {
+            cb(err_);
+            return;
+        }
+        var ref = dockerManifestRef || rat.tag;
+    client.getManifest({ref: ref}, function (err, dockerManifest, res) {
         if (err) {
             if (err.statusCode === 404) {
                 cb();
@@ -238,7 +243,11 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
             }
         );
 
-        var configDigest = dockerManifest.config.digest;
+        var configDigest = dockerManifest.config && dockerManifest.config.digest;
+        if(!configDigest){
+             cb(new errors.InvalidDockerInfoError('Missing config digest in manifest'));
+             return;
+        }
         client.createBlobReadStream({digest: configDigest},
             function (err, stream, ress) {
                 if (err) {
@@ -269,16 +278,16 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
     
                 // stream error handling
                 stream.on('error', function (err) {
-                   console.error('read stream error:', err);
-                   cb(err);
+                self.log.error('read stream error:', err);
+                    cb(err);
                 });
     
                 stream.resume();
             }
         );
     });
-}
-
+  });
+};
 
 DockerSource.prototype.titleFromImportInfo =
 function titleFromImportInfo(importInfo) {
--- a/src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
+++ b/src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
@@ -6,6 +6,7 @@
 
 /*
  * Copyright 2016 Joyent, Inc.
+ * Copyright 2025 Edgecast Cloud LLC.
  */
 
 /*
@@ -32,7 +33,6 @@ var common = require('./common');
 var DockerJsonClient = require('./docker-json-client');
 var errors = require('./errors');
 
-
 // --- Globals
 
 // https://github.com/docker/docker/blob/77da5d8/registry/config_unix.go#L10
@@ -42,7 +42,10 @@ var MEDIATYPE_MANIFEST_V2
     = 'application/vnd.docker.distribution.manifest.v2+json';
 var MEDIATYPE_MANIFEST_LIST_V2
     = 'application/vnd.docker.distribution.manifest.list.v2+json';
-
+var MEDIATYPE_OCI_MANIFEST_LIST_V1
+    = 'application/vnd.oci.image.index.v1+json';
+var MEDIATYPE_OCI_IMAGE_MANIFEST_V1
+    = 'application/vnd.oci.image.manifest.v1+json';
 
 // --- internal support functions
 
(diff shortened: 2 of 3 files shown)


src/img/lib/sources/docker.js
```diff
@@ -238,7 +243,11 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
             }
         );
 
-        var configDigest = dockerManifest.config.digest;
+        var configDigest = dockerManifest.config && dockerManifest.config.digest;
+        if(!configDigest){
```
danmcd: Style nit:  `if (!configDigest) {`

src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
```diff
@@ -1590,6 +1593,127 @@ RegistryClientV2.prototype.createBlobReadStream =
...
+                            manifest.manifests.length === 0) {
+                        cb(new restifyErrors.InvalidContentError(fmt(
+                            'no manifests in %s:%s manifest list',
+                            self.repo.localName, opts.ref)));
+                        return;
+                    }
+                    for (var i = 0; i < manifest.manifests.length;  i++) {
+                        var platform = manifest.manifests[i].platform;
+                        if (platform &&
+                            platform.architecture === 'amd64' &&
```
danmcd: I honestly cannot remember if we ONLY ever want amd64 of if i386 is okay as well?  I'm happy to be educated here.

danmcd (commented):
Pass 0 or 1, mostly nits, but the last big change will require pass 1/2.

ALSO => Do we need to version-bump any `package.json` files because of these changes?

src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
cneira: According to this document https://github.com/opencontainers/image-spec/blob/main/image-index.md, in our case we could support  '386' or 'amd64'. I have not tested with a 386 Docker image, I'll try to create a test case for that scenario.

cneira: I added a test case for this scenario in the ticket.

src/img/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "imgadm",
   "description": "Manage SmartOS virtual machine images.",
-  "version": "3.9.4",
+  "version": "3.9.5",
   "author": "MNX Cloud (mnxsolutions.com)",
```
danmcd: Edgecast (gonna be the first one we do here).

src/img/lib/sources/docker.js
```diff
@@ -244,7 +244,7 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
         );
 
         var configDigest = dockerManifest.config && dockerManifest.config.digest;
-        if(!configDigest){
+        if(!configDigest) {
```
danmcd: Still missing one space after 'if': `if (!configDigest) {`

src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
```diff
@@ -42,7 +42,10 @@ var MEDIATYPE_MANIFEST_V2
     = 'application/vnd.docker.distribution.manifest.v2+json';
 var MEDIATYPE_MANIFEST_LIST_V2
     = 'application/vnd.docker.distribution.manifest.list.v2+json';
-
+var MEDIATYPE_OCI_MANIFEST_LIST_V1
+    = 'application/vnd.oci.image.index.v1+json';
+var MEDIATYPE_OCI_IMAGE_MANIFEST_V1
```
danmcd: Is this used anywhere? It's defined here, added in a table down below, but I don't see anything handling it.

src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
```diff
@@ -1590,6 +1593,128 @@ RegistryClientV2.prototype.createBlobReadStream =
     });
 };
 
+/*
+ * Get a manifest digest from a MEDIATYPE_MANIFEST_LIST_V2
+ * that needs to be used to get the actual manifest object.
+ *
+ *   client.getManifest({ref: <tag>}, function (err, manifestRef, res)
```
danmcd: `getManifestRef` ?  `manifestRef` on the callback? Down below it says, `manifestDigest`.

cneira: I updated the naming on the function to be clearer on the purpose of this helper function.

src/img/node_modules/docker-registry-client/lib/registry-client-v2.js
```diff
@@ -1590,6 +1591,129 @@ RegistryClientV2.prototype.createBlobReadStream =
...
+ * MEDIATYPE_OCI_MANIFEST_LIST_V1 payload.
+ * The retrieved digest needs to be used by client.getManifest to
+ * retrieve the actual manifest object.
+ *
+ * client.getManifestDgst({ref: <tag>}, function (err, manifestDgst, res) {
+ *   });
+ * @param ref: {Object}
+ * @param cb {Function} `function (err, manifestDgst, res)`
+ */
+RegistryClientV2.prototype.getManifestDgst =function getManifestDgst(opts, cb) {
```
danmcd: NIT: spacing around `=`:  ` = function getManifestDgst(...`

src/img/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "imgadm",
   "description": "Manage SmartOS virtual machine images.",
-  "version": "3.9.4",
-  "author": "MNX Cloud (mnxsolutions.com)",
+  "version": "3.9.5",
+  "author": "Edgecast Cloud LLC",
```
danmcd: Nit: `"Edgecast Cloud LLC (edgecast.io)"`

src/img/lib/sources/docker.js
```diff
@@ -228,57 +345,69 @@ DockerSource.prototype.getImportInfo = function getImportInfo(opts, cb) {
...
-                        imgId: dockerManifest.layers[dockerManifest.layers.length-1].digest,
-                        imgJson: imgJson,
-                        dockerManifest: dockerManifest,
-                        layerDigests: layerDigests,
-                        uuid: imgmanifest.imgUuidFromDockerDigests(layerDigests)
-                    };
-                    cb(null, importInfo);
-                }));
-    
+                stream.pipe(
```
danmcd: Why the change in spacing?  Is this to be consistent with the other client? Or are there indentation changes I'm missing?

src/img/node_modules/docker-registry-client/lib/index.js
```diff
@@ -132,8 +132,11 @@ module.exports = {
     pingV2: reg2.ping,
     loginV2: reg2.login,
     digestFromManifestStr: reg2.digestFromManifestStr,
-    MEDIATYPE_MANIFEST_V2: reg2.MEDIATYPE_MANIFEST_LIST_V2,
+    MEDIATYPE_MANIFEST_V2: reg2.MEDIATYPE_MANIFEST_V2,
```
danmcd, line 136: Is this a longstanding bug?

cneira: It seems so, but as the exported value was not being used no harm was done.

src/img/lib/sources/docker.js
cneira: The callback was more than 80 columns, so I changed the indentation to fix that. I’ll double check the indentation that I applied.

cneira: The `stream.pipe` call matches the`if`indentation from the previous line, let me know if I need to change the indentation on this callback.

danmcd: So no code changed here *in this section*; just the indentation, right? Sure looks like that's the case. If so, please confirm so I can resolve this one.

src/img/lib/sources/docker.js
```diff
@@ -162,6 +161,126 @@ function _importInfoFromV21Manifest(opts, cb) {
...
+
+      if (!dockerManifest.config.digest) {
+        cb(
+          new errors.InvalidDockerInfoError(
+            "No docker manifest config defined"
+          )
+        );
+        return;
+      }
+      var configDigest = dockerManifest.config.digest;
```
danmcd, line 238: Is this setup and call of createBlobreadStream() the same as the one below?  IF SO can it be refactored into a single function both here and the original one below can call?  I may be missing subtleties, however.

danmcd (commented):
The createReadBlobStream() calls in two places has me wondering.  If I'm missing a sublety, please let me know.

src/img/lib/sources/docker.js
cneira: That's right, there are only indentation changes to that callback to enforce 80-column lines.

src/img/lib/sources/docker.js
cneira: You are right, it's the same. I encapsulated the manifest V2 list handling, but I did not factor out that common code section. I agree that less code is better. I'll factor that code section into a utility function.
Good catch!!.

src/img/lib/sources/docker.js
```diff
@@ -162,6 +161,142 @@ function _importInfoFromV21Manifest(opts, cb) {
...
+        cb(null, importInfo);
+      });
+    }
+  );
+}
+
+/*
+ * Converts a blob digest into a JSON object.
+ */
+function _getBlobAsJson(client, digest, cb) {
```
danmcd, line 266: Pardon my lack of JS knowledge, this function can appear after its first caller above?

danmcd (commented):
One stupid JS question...

src/img/lib/sources/docker.js
cneira: Yes, is ok to the function to appear after its first caller. In JavaScript this is called function hoisting.

PR #1153, opened 2025-07-31
OS-8680 Eliminate in-tree pwgen(1) in smartos-live

danmcd:
Subject says it all. Analysis and test notes will be in the ticket.


tsoome (approved):
Looks like OK, thanks.

configure
```diff
@@ -201,7 +202,7 @@ function install_pkgin
 {
 	local pkglist
 
-	pkglist="build-essential flex libxslt openjdk11 nodejs"
+	pkglist="build-essential flex libxslt openjdk11 nodejs pwgen"
```
cneira, line 205: I have a doubt, the list of packages is pinned to the specific versions provided by the platform image pkgsrc installed, right?. I'm assuming that the answer is yes.

danmcd: It's dependent on the *pkgsrc* version.  Our current requirement is that build zones must be at 24.4.1. This means whatever versions 24.4.1 suppports, we get. It's why we have official pkgsrc versions for building SmartOS.

It also means if the pkgsrc version of a tool is broken (like `pwgen` actually is in 24.4.1), we need to fix it either by making sure it works in that pkgsrc version (cf. TritonDataCenter/pkgsrc#405 ), or we do not use the tool until the next fixed pkgsrc version.

nshalman (approved):

configure
```diff
@@ -223,6 +224,11 @@ function install_pkgin
 		    "failed to install $pkg"
 		fi
 	done
+
+	#
+	# Make sure we've upgraded all of our installed packages.
+	#
+	$conf_priv pkgin -y upgrade || fatal "failed installed-package upgrades"
```
danmcd: @jperkin suggested I move this before any installations. I'm likely to move it to the beginning of this function.  Please confirm?

danmcd: Nope, I'm wrong: move it to after `update` but before installations.

jperkin (approved):

danmcd:
Testing notes are in the ticket!

cneira (approved):

danmcd:
Pushed manually due to weirdness with IA automation.

PR #1158, opened 2025-09-08
OS-8690: smartos-ui does not adapt to changed admin interface IP address

nshalman:
Fixes tritondatacenter/smartos-ui#19

With this change applied:
```
[root@smartos ~]# svcs vmadmd smartos-ui smartos-ui-executor
STATE          STIME    FMRI
online         19:06:31 svc:/system/smartdc/vmadmd:default
online         19:06:31 svc:/system/smartdc/smartos-ui-executor:default
online         19:06:31 svc:/system/smartdc/smartos-ui:default
[root@smartos ~]# dladm create-vnic -l ixgbe0 dhcptest0
[root@smartos ~]# ifconfig dhcptest0 plumb up
[root@smartos ~]# ifconfig dhcptest0 dhcp
[root@smartos ~]# svcs vmadmd smartos-ui smartos-ui-executor
STATE          STIME    FMRI
online         19:08:47 svc:/system/smartdc/smartos-ui-executor:default
online         19:08:47 svc:/system/smartdc/smartos-ui:default
online         19:08:47 svc:/system/smartdc/vmadmd:default
[root@smartos ~]# ifconfig dhcptest0 down
[root@smartos ~]# ifconfig dhcptest0 unplumb
[root@smartos ~]# dladm delete-vnic dhcptest0
```

### DIFF
--- a/src/etc/dhcp/eventhook
+++ b/src/etc/dhcp/eventhook
@@ -18,6 +18,14 @@
 # This only applies in the global zone!
 [[ $(zonename) == global ]] || exit 0
 
+# We don't care which interface. If any GZ IPs have changed
+# then we want to update the sysinfo cache.
+#INTERFACE=$1
+
+# We only support this for the BOUND/BOUND6 actions.
+ACTION=$2
+[[ "$ACTION" == BOUND* ]] || exit 0
+
 # Refresh the sysinfo cache
 sysinfo -u
 
--- a/src/etc/dhcp/eventhook
+++ b/src/etc/dhcp/eventhook
@@ -29,7 +29,8 @@ ACTION=$2
 # Refresh the sysinfo cache
 sysinfo -u
 
-# Restart any affected services that depend on sysinfo
+# Restart any affected services currently known to depend on binding to the
+# Admin_IP obtained from sysinfo.
 # (Note that if any of these services do not exist
 # then they just won't show up on stdout.)
 svcs -Ho fmri vmadmd smartos-ui smartos-ui-executor | while read -r service
--- a/src/etc/dhcp/eventhook
+++ b/src/etc/dhcp/eventhook
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 Edgecast Cloud LLC.
+#
+
+# This only applies in the global zone!
+[[ $(zonename) == global ]] || exit 0
+
+# Refresh the sysinfo cache
+sysinfo -u
+
+# Restart any affected services that depend on sysinfo
+# (Note that if any of these services do not exist
+# then they just won't show up on stdout.)
+svcs -Ho fmri vmadmd smartos-ui smartos-ui-executor | while read -r service
+do
+	svcadm restart "$service"
+done
--- a/src/Makefile
+++ b/src/Makefile
@@ -354,6 +354,7 @@ LIBSDC_TARGETS = \
 	lib/sdc/zone.sh
 
 ETC_TARGETS = \
+	etc/dhcp/eventhook \
 	etc/nodename \
 	etc/rtc_config \
 	etc/ssh/sshd_config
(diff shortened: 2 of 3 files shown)


src/etc/dhcp/eventhook
```diff
@@ -0,0 +1,30 @@
...
+#
+# Copyright 2025 Edgecast Cloud LLC.
+#
+
+# This only applies in the global zone!
+[[ $(zonename) == global ]] || exit 0
+
+# Refresh the sysinfo cache
+sysinfo -u
+
```
danmcd, line 31: Do we wish to do this blindly?  Or do we need to evaluate what event (new-IP) would require `sysinfo -u` ?  IIRC it can take a while, and if the event-hook executes before DHCP finishes its job we could sabotage DHCP.

danmcd: The `dhcpagent`(8) pages indicate a BOUND (and eventually BOUND6 IFF we use dhcpv6) event should be monitored, which could help use narrow our use of `sysinfo -u`.

```
       BOUND and BOUND6

           These events occur during interface configuration. The event
           program is invoked when dhcpagent receives the DHCPv4 ACK or DHCPv6
           Reply message from the DHCP server for the lease request of an
           address, indicating successful initial configuration of the
           interface. (See also the INFORM and INFORM6 events, which occur
           when configuration parameters are obtained without address leases.)
```

nshalman: It takes about 1 second on my machine.  I guess the question is how often do we think it will fire.
```
[root@smartos ~]# time sysinfo -u &>/dev/null

real    0m1.024s
user    0m0.407s
sys     0m0.678s
[root@smartos ~]# time sysinfo &>/dev/null

real    0m0.022s
user    0m0.008s
sys     0m0.013s
```

nshalman: But I think you're right. We should examine the arguments.

nshalman: please take a look.

danmcd: See above.

src/etc/dhcp/eventhook
```diff
@@ -0,0 +1,38 @@
...
+# This only applies in the global zone!
+[[ $(zonename) == global ]] || exit 0
+
+# We don't care which interface. If any GZ IPs have changed
+# then we want to update the sysinfo cache.
+#INTERFACE=$1
+
+# We only support this for the BOUND/BOUND6 actions.
+ACTION=$2
+[[ "$ACTION" == BOUND* ]] || exit 0
```
danmcd, line 27: Does wildcarding work in that situation? Should `BOUND*` be in quotations like `"$ACTION"` is ?

nshalman:
```
[root@smartos ~]# ACTION=BOUND6 ; [[ "$ACTION" == "BOUND*" ]] && echo works
[root@smartos ~]# ACTION=BOUND6 ; [[ "$ACTION" == BOUND* ]] && echo works
works
[root@smartos ~]# [[ "BOUND" == "BOUND*" ]] && echo "it could be in quotation marks"
[root@smartos ~]# [[ "BOUND" == "BOUND*" ]] || echo nope
nope
```

danmcd: Much, thank you!

src/etc/dhcp/eventhook
```diff
@@ -0,0 +1,38 @@
...
+
+# Refresh the sysinfo cache
+sysinfo -u
+
+# Restart any affected services that depend on sysinfo
+# (Note that if any of these services do not exist
+# then they just won't show up on stdout.)
+svcs -Ho fmri vmadmd smartos-ui smartos-ui-executor | while read -r service
```
danmcd, line 36: Suspenders for the belt of exit-if-NGZ, huh?  :)

And "depend on sysinfo"? Or "Depend on sysinfo's changing of `Admin_IP` ?

tsoome (approved):

cneira (approved):

src/etc/dhcp/eventhook
nshalman: This script doesn't even show up in the NGZs as we currently create them anyway. 
Also, in a NGZ `sysinfo -u` is a noisy no-op so I think we've got quite a few belts and suspenders:
```
[root@smartos-builds ~]# sysinfo -u ; echo $?
Error: This program can only be run in the global zone.
0
```
As for the code comment, I think those two are equivalent because these services get the `Admin_IP` FROM the `sysinfo` command rather than the cache file directly. Perhaps "Restart any affected services currently known to depend on binding to the Admin_IP obtained from sysinfo"??

Let me know if you do indeed want the comment tweaked.

nshalman:
```
OS-8690: smartos-ui does not adapt to changed admin interface IP address

Reviewed by: Carlos Neira <carlos.neira@edgecast.io>
Reviewed by: Dan McDonald <danmcd@edgecast.io>
Reviewed by: Toomas Soome <tsoome@edgecast.io>
Approved by: Dan McDonald <danmcd@edgecast.io>
```

src/etc/dhcp/eventhook
danmcd: > Perhaps "Restart any affected services currently known to depend on binding to the Admin_IP obtained from sysinfo"??
> 
> Let me know if you do indeed want the comment tweaked.

Please use the exact text you proposed here, I like it.

danmcd:
Given @cneira 's approval was on code that only is altered by a comment rewording, I think he'll be okay.  If he doesn't re-+1 this in 24 hours, I'll turn on IA anyway. If he does re-+1 sooner, I'll turn it on ASAP.

cneira (approved):

PR #1161, opened 2025-10-18
OS-8698 smartos-live ucodecheck needs to address AMD fallback ucode

danmcd:


danmcd:
Testing notes in ticket, and a Jenkins build with a pre-17638 will be running shortly, see the `continuous-integration` links.

danmcd:
After several whacks, the works before-and-after `ucodecheck` tool in smartos-live is working properly both before, and after, the next AMD ucode drop.

cneira (approved):

PR #1162, opened 2025-11-06
OS-8701 smartos-live (and node-0.10 of node-fs-ext) need to freeze "nan" version

danmcd:
Needs from-scratch build testing and also testing on Jenkins.

This may be the tip of an iceberg.  If it spreads into Triton, we'll have to open up tickets there.


travispaul (approved):

danmcd:
Built fine, running on two machines, including one that's building SmartOS as I type this.

PR #1166, opened 2025-11-19
TRITON-2520: Servers should not change UUID on reboot

nshalman:
```
Portions generated by: Claude <noreply@anthropic.com>
```

Code review of the current version by Claude:

```mermaid
flowchart TD
    Start([get_or_store_uuid called]) --> CheckZpool{Is $Zpool empty?}

    CheckZpool -->|Yes| Return([Return - do nothing])
    CheckZpool -->|No| GetSource[Get ZFS property source for<br/>org.smartos:server_uuid<br/>on $Zpool/var]

    GetSource --> CheckSource{Is source == 'local'?}

    CheckSource -->|No| NoStoredUUID[No stored UUID on pool]
    CheckSource -->|Yes| ReadStored[Read UUID value from pool]

    ReadStored --> CheckOverride{Is OVERRIDE_UUID set<br/>AND different from<br/>stored value?}

    CheckOverride -->|Yes| ValidateOverride{Is OVERRIDE_UUID<br/>a valid UUID?}
    CheckOverride -->|No| ValidateStored{Is stored UUID<br/>valid?}

    ValidateOverride -->|Yes| UpdatePool[Update pool with<br/>OVERRIDE_UUID]
    ValidateOverride -->|No| ValidateStored

    UpdatePool --> ReadBack[Read updated value<br/>back from pool]
    ReadBack --> ValidateStored

    ValidateStored -->|Yes| UseStored[Set UUID to stored value]
    ValidateStored -->|No| CleanupInvalid[Delete invalid UUID<br/>from pool using<br/>zfs inherit]

    UseStored --> End([Return])

    CleanupInvalid --> NoStoredUUID
    NoStoredUUID --> ValidateCurrent{Is current UUID<br/>valid?}

    ValidateCurrent -->|Yes| PersistUUID[Store current UUID<br/>on pool]
    ValidateCurrent -->|No| End

    PersistUUID --> End
```

## UUID Priority Order

The function implements the following priority for UUID sources:

1. **Highest Priority**: `override_uuid` bootparam (stored in `$OVERRIDE_UUID`)
   - If present, updates the pool and uses this value
2. **Second Priority**: `org.smartos:server_uuid` on `$Zpool/var`
   - Used if valid and no override is present
3. **Fallback**: `$UUID` from smbios
   - Used only if nothing valid is stored on pool

## Key Decision Points

- **Empty Zpool check**: Early exit if no pool is available
- **Source check**: Determines if a UUID has been previously stored
- **Override handling**: OVERRIDE_UUID takes precedence and updates pool if different
- **Validation**: All UUIDs are validated using `valid_uuid()` helper before use
- **Cleanup**: Invalid UUIDs are removed from pool to prevent corruption

## Side Effects

- May update `$UUID` global variable to the pool-stored value
- May write to ZFS pool property `org.smartos:server_uuid`
- May delete invalid pool property values

### DIFF
--- a/src/sysinfo
+++ b/src/sysinfo
@@ -1191,6 +1191,8 @@ function valid_uuid(){
 # 3. org.smartos:server_uuid on $Zpool
 # We prefer 3 over 1
 # We prefer 2 over 3 and update 3 if they disagree.
+# NOTE: This function is only effective if called
+# AFTER both get_zpool and get_bootparams.
 function get_or_store_uuid()
 {
     [[ -z "$Zpool" ]] && return
@@ -1233,6 +1235,7 @@ if [[ ${ZONENAME} == "global" ]]; then
     get_disks
     get_recovery_configs
     get_bootparams
+    get_or_store_uuid
     get_psrinfo
     get_sdc_version
     if [[ -n ${SMARTDC_VERSION} ]]; then
@@ -1258,7 +1261,6 @@ get_smartos_vnics
 get_aggr_params
 get_boot_time
 get_dc_info
-get_or_store_uuid
 
 # whenever we update the cache, update both
 if [[ ${CACHE} == "true" ]]; then


src/sysinfo
```diff
@@ -1175,6 +1176,47 @@ fi
...
+    local uuid_source=$(zfs get -Hp -o source ${key} ${Zpool})
+    if [[ "$uuid_source" != "local" ]]; then
+        # Nothing recorded on the pool. Trust whatever we have
+        # if it is valid
+        valid_uuid ${UUID} || return
+        zfs set ${key}=${UUID} ${Zpool}
+    else
+        # Read what was stored on the pool.
+        local uuid_value=$(zfs get -Hp -o value ${key} ${Zpool})
+        valid_uuid ${uuid_value} || return
```
nshalman: We shouldn't return here. We might be able to clean this up with an override in a moment...

src/sysinfo
```diff
@@ -1175,6 +1176,47 @@ fi
...
+function get_or_store_uuid()
+{
+    [[ -z "$Zpool" ]] && return
+    local key="com.tritondatacenter:uuid"
+    local uuid_source=$(zfs get -Hp -o source ${key} ${Zpool})
+    if [[ "$uuid_source" != "local" ]]; then
+        # Nothing recorded on the pool. Trust whatever we have
+        # if it is valid
+        valid_uuid ${UUID} || return
+        zfs set ${key}=${UUID} ${Zpool}
```
danmcd: So `com.tritondatacenter:uuid` on the *dataset* ${Zpool}, vs. ${Zpool}/var like we do with `sdc-factoryreset` ?  I'm thinking we might wish to keep it in `/var` in case we see this during an installation or setup process and it happens too-early? Or maybe too-early is what we're trying to avoid?

src/sysinfo
```diff
@@ -1175,6 +1176,45 @@ fi
...
+# it on subsequent boots.
+# Sources of UUID:
+# 1. `smbios`
+# 2. override_uuid bootparam
+# 3. com.tritondatacenter:uuid on $Zpool
+# We prefer 3 over 1
+# We prefer 2 over 3 and update 3 if they disagree.
+function get_or_store_uuid()
+{
+    [[ -z "$Zpool" ]] && return
```
danmcd, line 1198: One more philosophical question: Should we ONLY do this on headnode and CN (i.e. `bootparams | grep -q smartos=true` fails) ?

danmcd (commented):
Forgot to mention: Copyright update please.

src/sysinfo
nshalman: I don't think so. I don't see any harm in this on a SmartOS node. I have no problem putting this in place on my machine, etc.

src/sysinfo
nshalman: Moved to `/var` dataset. Thanks!

src/sysinfo
danmcd: I see you already marked the bug as public, this is good.  When we detail this in the announcement email we need to cite the bug, its test notes, and the answer to why-all?

Resolving.

tsoome (approved):

src/sysinfo
nshalman: Also, the attribute has been renamed to `org.smartos:server_uuid` which makes it less odd to be shipping in SmartOS.

src/sysinfo
```diff
@@ -1175,6 +1177,55 @@ fi
...
+        # If we received a valid override, update the pool if needed.
+        if [[ -n ${OVERRIDE_UUID} && ${OVERRIDE_UUID} != ${uuid_value} ]]; then
+            valid_uuid ${OVERRIDE_UUID} && \
+                zfs set ${key}=${OVERRIDE_UUID} ${dataset}
+            # Explicitly read it back in.
+            uuid_value=$(zfs get -Hp -o value ${key} ${dataset})
+        fi
+        if valid_uuid ${uuid_value}; then
+            # Pool contains a valid UUID. Use it.
+            UUID=${uuid_value}
```
danmcd, line 1215: Upon further review, do we want to complain somewhere if the existing `UUID` is not the same as `uuid_value` for diagnostic purposes? I don't think there's a wrong answer here beyond, "no idea." Given how frequently this gets called, maybe not?

src/sysinfo
```diff
@@ -1207,6 +1258,7 @@ get_smartos_vnics
 get_aggr_params
 get_boot_time
 get_dc_info
+get_or_store_uuid
```
danmcd: Also, should this go in the GZ-only section above (say near line 1237 after get_or_store_uuid)?

danmcd (commented):
Two last-minute after-further-review gotchas.

src/sysinfo
nshalman: Very good catch. Thanks!

src/sysinfo
nshalman: The main question in my mind is where would one expect to find any non-fatal message from sysinfo in the first place? I can emit something, but I don't know how useful it would be. Also, that would only get tripped on the sysinfo calls that are (re)generating the cache...
This is already way more validation than was ever done here, and it's all just defensive in my opinion.

src/sysinfo
```diff
@@ -1191,6 +1191,8 @@ function valid_uuid(){
 # 3. org.smartos:server_uuid on $Zpool
 # We prefer 3 over 1
 # We prefer 2 over 3 and update 3 if they disagree.
+# NOTE: This function is only effective if called
+# AFTER both get_zpool and get_bootparams.
```
danmcd, line 1195: NICE!

src/sysinfo
danmcd: You've made a good case for no-need.  Resolving.

danmcd (approved):
Much better.

tsoome (approved):

src/sysinfo
danmcd: Resolving, and this text here is present for future expansion if we change this from "implementation-detail/unstable-interface."

PR #1175, opened 2026-03-06
OS-8721 SmartOS builds in Jenkins should guarantee different buildstamps per stage

danmcd:
First full build is running now.  Once I got over the stupid shell mistakes I made (variable name mismatch e.g.) things appear to be building.  The output stamps in manta's `platform{,-debug,-gcc14}` entries will be the {dis,}proof in the pudding.

### DIFF
--- a/tools/build_jenkins
+++ b/tools/build_jenkins
@@ -8,6 +8,7 @@
 #
 # Copyright 2022 Joyent, Inc.
 # Copyright 2022 MNX Cloud, Inc.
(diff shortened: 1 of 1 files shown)


danmcd:
Initial testing notes are on the ticket, and subsequent requests tests will be there too.

danmcd:
An additional testing note, easier to confirm, is now in the ticket.

travispaul (approved):

PR #1176, opened 2026-03-19
OS-8723 Expose virtio1 zone attribute via vmadm

danmcd:
A possibility...

### DIFF
--- a/src/vm/man/vmadm.8.md
+++ b/src/vm/man/vmadm.8.md
@@ -2126,26 +2126,13 @@ tab-complete UUIDs rather than having to type them out for every command.
         update: yes
         default: 'std'
 
-    virtio09:
-
-	Enables or disables the presentation of virtio 0.9 devices to the
-	guest of any sort. Normally a transitional device (0.9 and 1.0) is
-	presented to the guest, but some guest testing may require hiding
-	0.9 support. Defaults to true.
-
-	type: boolean
-	vmtype: bhyve
-	listable: yes
-	create: yes
-	update: yes
-	default: true
-
     virtio1:
 
-	Enables or disables the presentation of virtio 1.0 devices to the
-	guest of any sort. Some guests have drivers that do not support
-        the transitional (both 0.9 and 1.0) interface for 0.9 devices due
-	to guest driver bugs, so this defaults to false.
+	Enables or disables the presentation of virtio 1.0 devices of any
+	sort to a bhyve guest. Some guests have drivers that do not support
+	the transitional (both 0.9 and 1.0) interface for 0.9 devices due to
+	guest driver bugs, so this defaults to false, for now. The default
+	may change in a future SmartOS version.
 
 	type: boolean
 	vmtype: bhyve
--- a/src/vm/man/vmadm.8.md
+++ b/src/vm/man/vmadm.8.md
@@ -2128,18 +2128,18 @@ tab-complete UUIDs rather than having to type them out for every command.
 
     virtio1:
 
-	Enables or disables the presentation of virtio 1.0 devices of any
-	sort to a bhyve guest. Some guests have drivers that do not support
-	the transitional (both 0.9 and 1.0) interface for 0.9 devices due to
-	guest driver bugs, so this defaults to false, for now. The default
-	may change in a future SmartOS version.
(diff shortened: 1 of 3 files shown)


src/vm/node_modules/VM.js
```diff
@@ -12680,6 +12680,7 @@ exports.create = function (payload, options, callback)
             }
             cb();
         }, function (cb) {
+	    // XXX KEBE ASKS, put a `virtio=true` here?
             normalizePayload(payload, null, log, function (err) {
```
cneira, line 12691: I think this is right, any mutation on the `virtio1` property should be done in `normalizePayload`, seems this is the pattern looking at the` docker.payload` code path, any patching on payload property or if that property turn on more flags/dependencies seems to be done here.

src/vm/man/vmadm.8.md
```diff
@@ -2126,6 +2126,34 @@ tab-complete UUIDs rather than having to type them out for every command.
...
+	presented to the guest, but some guest testing may require hiding
+	0.9 support. Defaults to true.
+
+	type: boolean
+	vmtype: bhyve
+	listable: yes
+	create: yes
+	update: yes
+	default: true
+
```
nshalman: Based on the other changes, this section should also be removed??

src/vm/node_modules/proptable.js
```diff
@@ -2080,6 +2080,16 @@ exports.properties = {
...
+    }, virtio1: {
+        loadValueTranslator: 'utils.fixBoolean',
+        payload: {
+            allowed: {
+                'bhyve': ['create', 'receive', 'update'],
+            },
+            type: 'boolean',
+            default: false
+        },
+        zonexml: 'zone.attr.virtio1'
```
nshalman: My concern here is that upon new VM creation, this will cause a new VM to get the property set to "false" which is not null and will cause my quick fix work to actually default to removing the override of modern=false and cause the effect to be "true".

src/vm/man/vmadm.8.md
danmcd: Yes.

src/vm/node_modules/proptable.js
danmcd: We're fixing that in VM.js.

src/vm/man/vmadm.8.md
```diff
@@ -2126,6 +2126,21 @@ tab-complete UUIDs rather than having to type them out for every command.
...
+	the transitional (both 0.9 and 1.0) interface for 0.9 devices due to
+	guest driver bugs, so this defaults to false, for now. The default
+	may change in a future SmartOS version.
+
+	type: boolean
+	vmtype: bhyve
+	listable: yes
+	create: yes
+	update: yes
+	default: false
```
nshalman: so this should now be `default: true`, correct?

danmcd: This is where it gets tricky: is "default" for creation? Or for if-no-value-in-object?

Oooh, I found a good prior example:  `autoboot`.  Maybe we should not specify the default value here at all, like `autoboot` does, and document better in the description?

src/vm/node_modules/proptable.js
danmcd: And maybe we should nuke `default`  here as well, per your comment in VM.js.

src/vm/man/vmadm.8.md
```diff
@@ -2126,6 +2126,21 @@ tab-complete UUIDs rather than having to type them out for every command.
         update: yes
         default: 'std'
 
+    virtio1:
+
+	Enables or disables the presentation of virtio 1.0 devices in the
+	transitoinal driver to a bhyve guest. Some guests have drivers that
```
nshalman: typo. should be "transitional"

danmcd:
Testing notes are in the ticket.  We will need more testing in this portion.

danmcd:
I should probably set up a PI with this on one Kebecloud CN, and a PI without this on another, and see what happens when `sdc-migrate` of a BHYVE VM comes into play.

danmcd:
Test results added to ticket.

nshalman (approved):

cneira (approved):

danmcd:
I will try and make sure the release notes cover this thoroughly.

PR #1177, opened 2026-03-30
OS-8724 - Reserve vendor-data for SmartOS/Triton use

travispaul:
Pull vendor-data from internal_metadata instead of customer_metadata.

Cloud-init will merge user-data and vendor-data (if the latter exists) preferring user-data: so there is no need for users to set vendor-data in customer_metadata. Moving vendor-data to internal_metadata makes this explicit and reserved for future platform use.

Testing notes in ticket.

### DIFF
--- a/src/vm/node_modules/cloudinit/nocloud.js
+++ b/src/vm/node_modules/cloudinit/nocloud.js
@@ -218,10 +218,10 @@ function userDataConfig(payload) {
 }
 
 function vendorDataConfig(payload) {
-    var customerMdata = payload.set_internal_metadata
+    var internalMdata = payload.set_internal_metadata
                         || payload.internal_metadata || {};
-    if (customerMdata.hasOwnProperty('cloud-init:vendor-data')) {
-        return customerMdata['cloud-init:vendor-data'];
+    if (internalMdata.hasOwnProperty('cloud-init:vendor-data')) {
+        return internalMdata['cloud-init:vendor-data'];
     }
     return undefined;
 }
--- a/src/vm/node_modules/cloudinit/nocloud.js
+++ b/src/vm/node_modules/cloudinit/nocloud.js
@@ -218,8 +218,8 @@ function userDataConfig(payload) {
 }
 
 function vendorDataConfig(payload) {
-    var customerMdata = payload.set_customer_metadata
-                        || payload.customer_metadata || {};
+    var customerMdata = payload.set_internal_metadata
+                        || payload.internal_metadata || {};
     if (customerMdata.hasOwnProperty('cloud-init:vendor-data')) {
         return customerMdata['cloud-init:vendor-data'];
     }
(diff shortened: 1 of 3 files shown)


danmcd:
Do you want this landing this week?  Or first-commit of next SmartOS?

danmcd:
Also, will this cause flag-days? Feels like it might, but I could use some context or more time to figure it out myself.  (More urgent if the answer to my first question is, "this week, please.")

travispaul:
@danmcd not critical for this week, but nice to have. I'm not 100% sure if this needs a flag day.
In the previous release users could set cloudinit data in vendor-data from customer_metadata, they can no longer do so. SmartOS users can still set vendor-data in internal_metadata if they wish but this won't be permitted in Triton from CloudAPI.

danmcd (commented):
One question:

If I have a VM object with both customer_metadata and internal_metadata:

```
# vmadm get $UUID | json uuid name alias customer_metadata internal_metadata
0b27d4ad-068e-417a-9907-6a923a3bc945
legacy-test
{
  "root_authorized_keys": "ssh-ed25519 <public-key> danmcd@mnx.io"
}
{
  "root_pw": "CENSORED"
}
# 
```

Will those be handled differently by a guest after this lands?  Or does it depend on the guest? Or am I overthinking this?  As of this moment both land in my guest but it's an old image (name kinda gives it away).

Otherwise this all seems sane.

danmcd (dismissed):
Read the docs; shouldn't be a problem.

travispaul:
> Will those be handled differently by a guest after this lands? Or does it depend on the guest?

In your example above there would be no change to the guest behavior.

The only way that this would impact behavior is if a user was previously setting `customer_metadata["cloud-init:vendor-data"]` to something in a VM object payload. Now, that value is ignored and is only honored if set in `internal_metadata["cloud-init:vendor-data"]`

src/vm/node_modules/cloudinit/nocloud.js
```diff
@@ -218,8 +218,8 @@ function userDataConfig(payload) {
 }
 
 function vendorDataConfig(payload) {
-    var customerMdata = payload.set_customer_metadata
-                        || payload.customer_metadata || {};
+    var customerMdata = payload.set_internal_metadata
+                        || payload.internal_metadata || {};
     if (customerMdata.hasOwnProperty('cloud-init:vendor-data')) {
         return customerMdata['cloud-init:vendor-data'];
     }
```
nshalman: This is fine. I can see that it works. It might be less confusing to s/customer/internal/ on the variable names?

nshalman (dismissed):
I am okay with this landing as is, I am also okay with the minor variable name improvement I mentioned.

src/vm/node_modules/cloudinit/nocloud.js
danmcd: @travispaul how painful would `s/customerMdata/internalMdata/g` be?

travispaul: That's a trivial change and I just pushed it up. Also ran `make check` and the nocloud unit tests after the change as a sanity test all good.

danmcd (approved):
Thank you, and I know @nshalman was okay EITHER way, so I'm IA-ing this and you should include both of us as reviewers.

PR #1179, opened 2026-05-10
vmadm create: reject invalid brand cleanly (#1168)

nwilkens:
Fixes #1168.

`normalizePayload()` runs `applyZoneDefaults()` (and `fixPayloadMemory()`) before the existing invalid-brand check. Both index `BRAND_OPTIONS[brand].features.*` unguarded, so an unsupported brand throws `TypeError: Cannot read property 'features' of undefined` and dumps core instead of returning the clean error path.

- Move the brand-existence check above the apply-defaults branch in `normalizePayload`. Explicit unsupported brands now return `Error: Unsupported brand: <name>` via callback. No-brand and valid-brand paths are unchanged.
- Add an `assert` in `fixPayloadMemory` so future call paths that skip upfront validation fail with a clear message rather than silently TypeError'ing on one of the ~9 unguarded `BRAND_OPTIONS[brand]` dereferences in this file.
- Regression test in `test-create.js` calls `VM.create({brand: "smartos", ...})` and asserts the callback gets `Unsupported brand`.

### DIFF
--- a/src/vm/node_modules/VM.js
+++ b/src/vm/node_modules/VM.js
@@ -4692,6 +4692,9 @@ function fixPayloadMemory(payload, vmobj, log)
         brand = payload.brand;
     }
 
+    assert(BRAND_OPTIONS.hasOwnProperty(brand),
+        'fixPayloadMemory: unsupported brand: ' + brand);
+
     if (BRAND_OPTIONS[brand].features.default_memory_overhead
         && payload.hasOwnProperty('ram')
         && !payload.hasOwnProperty('max_physical_memory')) {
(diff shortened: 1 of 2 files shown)


danmcd:
Probably beats the crud out of this (where I took a SmartOS UI json and `s/joyent/smartos/g`):

```
[root@alder-lake /var/crash/volatile]# vmadm create -f /tmp/gotcha.json 
Uncaught TypeError: Cannot read property 'features' of undefined

FROM
fixPayloadMemory (/usr/vm/node_modules/VM.js:4695:29)
applyZoneDefaults (/usr/vm/node_modules/VM.js:4977:5)
normalizePayload (/usr/vm/node_modules/VM.js:10083:9)
vasync.waterfall.vm_type (/usr/vm/node_modules/VM.js:12691:13)
Object._onImmediate (/usr/vm/node_modules/vasync/lib/vasync.js:875:12)
processImmediate [as _immediateCallback] (timers.js:330:15)
Abort (core dumped)
[root@alder-lake /var/crash/volatile]# 
```

Make sure you file an OS ticket for this, too.

PR #1180, opened 2026-05-11
OS-8735 piadm(8) lies about what version of loader is installed

danmcd:
Testing notes (so far) in ticket.  One last thing I'm a tiny bit worried about, but I also want to see if anyone here catches it in review.  The change thus far is straightforward.

### DIFF
--- a/src/piadm.sh
+++ b/src/piadm.sh
@@ -824,11 +824,13 @@ update_boot_sectors() {
 			# will get set.
 		else
 			# Plow through devices, even if some fail.
-			# installboot also does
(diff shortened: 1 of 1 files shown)


tsoome (approved):

tsoome (approved):

danmcd:
I was worried about disks without UEFI ESPs on them, but that is not going to be a problem *most likely*.  I tried one on a VMware VM I'd sorta set-aside, but it actually was a good test of no-ESP pools.

PR #1181, opened 2026-06-30
OS-8754 let Jenkinsfile skip stages in case of single-stage failure

danmcd:
At the very least the skip-stage booleans work.  Running a "regular build" now to make sure default behavior continues to work.

Going to see if I can get failed-stage to not kill the whole `parallel` stage too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,7 +17,6 @@ pipeline {
     options {
         buildDiscarder(logRotator(numToKeepStr: '30'))
         timestamps()
-        parallelsAlwaysFailFast()
     }
     // Don't assign a specific agent for the entire job, in order to better
     // share resources across jobs. Otherwise, we'd tie up an agent here for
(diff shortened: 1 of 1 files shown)


danmcd:
Testing notes are in the ticket.

Jenkinsfile
```diff
@@ -159,14 +164,15 @@ set -o pipefail
                 always {
                     cleanWs cleanWhenSuccess: true,
                         cleanWhenFailure: false,
-                        cleanWhenAborted: true,
+                         cleanWhenAborted: true,
```
tsoome: is there extra space added?

danmcd: Eeesh.... I should fix that to not be a diff.  Good catch.

tsoome (approved):

danmcd:
Testing notes are in the ticket.

PR #1182, opened 2026-08-04
OS-8759 Update pkgsrc-setup to 20260804

danmcd:
This *should* fix the fresh-installation case. As mentioned in the ticket, I'm curious if we should have an additional modification (somewhere) that updates things per the cited email. I've performed the steps in said email on my NUCs with success.  I will report a fresh ISO build's installation on the ticket after building.

### DIFF
--- a/src/smartdc/bin/pkgsrc-setup.sh
+++ b/src/smartdc/bin/pkgsrc-setup.sh
@@ -6,6 +6,7 @@
 #
 # Copyright 2022 Joyent, Inc.
 # Copyright 2024 MNX Cloud, Inc.
+# Copyright 2026 Edgecast Cloud LLC.
 #
 
 function fatal () {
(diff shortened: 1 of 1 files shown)


jperkin (approved):
I wouldn't bother with the legacy fix-up, I think there are too many potential hazards and added complications.

danmcd:
> I wouldn't bother with the legacy fix-up, I think there are too many potential hazards and added complications.

When we send the release email it should cite your email about how to update on an existing one, as all GZ-tools-pkgsrc deployments will run into this unless they, like your email says, updated pkgsrc in the GZ with some modicum of frequency.

danmcd:
Test notes now in ticket.  Can perform more testing upon request.

nshalman (approved):

PR #1184, opened 2026-08-31
OS-8766 config.sh performance improvement

danmcd:


nshalman (dismissed):
Looks good to me.

src/lib/sdc/config.sh
danmcd: Add `# Copyright 2026 Edgecast Cloud LLC` please.

nwilkens: Done. Replaced the MNX line with `# Copyright 2026 Edgecast Cloud LLC.`

src/lib/sdc/config.sh
```diff
@@ -8,7 +8,7 @@
 # bash config.sh -json
 #
 # Copyright 2018 Joyent Inc.
-# Copyright 2025 MNX Cloud, Inc.
+# Copyright 2026 Edgecast Cloud LLC.
```
danmcd, line 11: I would've ADDED the EC one.  Good to know that for history, even if staff is the same.  NOt a hill I need to die on at this time, however.

tsoome (approved):

